PR #145, opened 2022-10-21
feat: Pathwar - Cybersecurity Learn2Earn Platform v0.1

RedaDev17:
Update:

Bridge code is here: https://github.com/TERITORI/pathwar-bridge 

#### Technical 

 Quest |  How to verify
-- | --
Install and write your first realm | A simple query to the BC can see the address deployed
Get your first gno | A simple query can see bank/balance
First PR merged | Harder, as we will need to monitor GitHub, but there is an API.
Port a existing contract | Maybe the user can provide both source codes or both addresses of the contracts.
Hello world - finish tutorial | Get a realm rendered on the screen using gnoweb. (Easy (markdown) but not trivial, you need to have all the software dev tools
Try GnoLand IDE | Not sure how to verify
  |  
  |  

</body>
</html>

<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.01//EN" "http://www.w3.org/TR/html4/strict.dtd">
<html>
<head>

#### Everyone track:
<body>

 Quest |  How to verify
-- | --
Attend a gno call | How to verify?
Attend an event | Not sure how to verify
Get your first gno | A simple query can see bank/balance

</body>
</html>


omniwired:
More resources here

https://twitter.com/danielmakelley/status/1681678181055569920?s=46

omniwired:
https://twitter.com/voldimmoral/status/1686072531147493376?s=46

PR #384, opened 2022-12-15
feat/multisig-wallet

YoLoADR:
Obsolete.

Please use https://github.com/TERITORI/teritori-dapp/tree/feat/multisig-wallet


zxxma:
TODO:
- Enforce to add the multisig creator wallet adress by default at creation (but can be deleted)

omniwired:
closing not to be confused with the new one.

PR #440, opened 2023-01-24
fix(indexer): ignore p2e score update if problem

n0izn0iz:
this is deployed currently

do not merge, we might want to replay the chain with normal code


n0izn0iz:
no longer used

PR #442, opened 2023-01-31
feat: dual squad staking

hthieu1110:
- Add dual squad on UI
- Update indexer to handle user's multi squads at the same time 
- Update script p2e_update_score to handle multi squads


api/p2e/v1/p2e.proto
```diff
@@ -17,7 +17,7 @@ message AllSeasonsRequest {
 message SeasonWithoutPrize {
   string id = 1;
   string boss_name = 2;
-  int32 boss_hp = 3;
+  float boss_hp = 3;
```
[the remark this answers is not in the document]
hthieu1110: I've just recheck, it would be better to keep int32 here because it's used to calculate daily prize. Let's me change it

go/internal/indexerdb/nft.go
```diff
@@ -47,7 +46,7 @@ type NFT struct {
 	ImageURI    string
 	OwnerID     UserID
 	IsListed    bool
-	PriceAmount sql.NullString `gorm:"type:numeric"`
+	PriceAmount string
```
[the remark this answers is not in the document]
hthieu1110: I don't use this field directly in this PR so I will revert this, we will see it later. I need this mainly for ETH indexer PR

go/pkg/marketplace/marketplace_service.go
```diff
@@ -346,7 +346,7 @@ func (s *MarkteplaceService) NFTs(req *marketplacepb.NFTsRequest, srv marketplac
 			NetworkId:          nft.Collection.NetworkId,
 			ImageUri:           ipfsutil.IPFSURIToURL(imageURI),
 			IsListed:           nft.IsListed,
-			Price:              nft.PriceAmount.String,
+			Price:              nft.PriceAmount,
```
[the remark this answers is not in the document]
hthieu1110: You're right. It's not related to this PR directly. We will see it later

packages/screens/RiotGame/component/GameContentView.tsx
```diff
@@ -24,6 +25,7 @@ export const GameContentView: React.FC<GameContentViewProps> = ({
   containerStyle,
   contentStyle,
   bgImage,
+  bgImageURI,
```
n0izn0iz, line 28: why create a second prop and not pass `{ uri: bgImageURI }` at call sites?

packages/screens/RiotGame/component/FightCountdownSection.tsx
```diff
@@ -1,130 +1,159 @@
...
-  stakingState,
   unstake,
+  isUnstaking,
+  squad,
+  now,
+  cooldown,
 }) => {
+  const { remainingTime, stakingState } = useMemo(() => {
+    let _remainingTime = 0;
+    let _stakingState = StakingState.UNKNOWN;
```
n0izn0iz, line 49: why _ prefix?

packages/screens/RiotGame/RiotGameFightScreen.tsx
```diff
@@ -1,201 +1,135 @@
...
-      navigation.replace("RiotGameEnroll");
-    }
-  }, [isSquadLoaded, isStakingStateLoaded]);
-
-  // Start the timer
-  useEffect(() => {
-    if (!isSquadLoaded || !squadStakingConfig?.owner || !isLastStakeTimeLoaded)
-      return;
+    const interval = setInterval(() => {
+      setNow(+new Date());
```
n0izn0iz, line 50: `Date.now()` achieves the same thing and is less cryptic

go/internal/indexerdb/p2e.go
```diff
@@ -3,8 +3,10 @@ package indexerdb
 // Delete the record when user unstake
 type P2eSquadStaking struct {
 	OwnerID   UserID `gorm:"primaryKey"`
-	StartTime uint64 `gorm:"index"`
+	StartTime uint64 `gorm:"primaryKey"`
+	SeasonID  string `gorm:"index"`
 	EndTime   uint64 `gorm:"index"`
+	TokenIDs  string
```
n0izn0iz, line 9: shouldn't the primary key also include the `TokenIDs`?
this can break if the `StartTime` is the same for two squads, no?

hthieu1110: From our dapp, with 1 ownerID we cannot have the same StartTimes for 2 different stakes but theorically a malicious user can build the messages him self and send 2 stakes in the same block so you're right. I will add TokenIDs in compound keys to be sure.

packages/screens/RiotGame/RiotGameFightScreen.tsx
hthieu1110: Yep, use Date.now() instead of +new Date()

packages/screens/RiotGame/component/FightCountdownSection.tsx
hthieu1110: It was used to differentiate with remainingTime, stakingState from parent scope just to make the code more readable but I can remove _, there is no problem.

packages/screens/RiotGame/component/GameContentView.tsx
hthieu1110: You are totally right. Removed unneeded bgImageURI

n0izn0iz (approved):
LGTM

PR #455, opened 2023-02-12
feat: p2e season 3

hthieu1110:
- [x] update script p2e-update-leaderboard to reset score every day
- [x] add squad preset
- [x] fix twitter share post
- [x] add child collection to p2e marketplace


packages/screens/RiotGame/component/CollectionView.tsx
```diff
@@ -0,0 +1,45 @@
+import React, { useState } from "react";
+import { ScrollView } from "react-native-gesture-handler";
+
+import { SortDirection } from "../../../api/marketplace/v1/marketplace";
+import { useCollectionInfo } from "../../../hooks/useCollectionInfo";
+import {
+  TabsListType,
+  Header as CollectionHeader,
+  Content as CollectionContent,
+} from "../../Marketplace/CollectionScreen";
```
n0izn0iz, line 10: avoid importing across screens, better to move these components in components package

packages/screens/RiotGame/component/RiotGameHeader.tsx
```diff
@@ -35,7 +35,9 @@ import { PickByValue } from "../../../utils/types/helper";
 type MenuItem = {
   id: string;
   name: string;
-  route?: keyof PickByValue<RootStackParamList, undefined>;
+  route?:
+    | keyof PickByValue<RootStackParamList, undefined>
+    | "RiotGameMarketplace";
```
n0izn0iz, line 40: why do you need this? the `RiotGameMarketplace` should already be in the `keyof PickByValue<RootStackParamList, undefined>` if the navigation is configured correctly

packages/screens/RiotGame/component/RipperSelectorModalV2.tsx
```diff
@@ -21,7 +21,7 @@ import FlexRow from "../../../components/FlexRow";
 import { SVG } from "../../../components/SVG";
 import { TertiaryBox } from "../../../components/boxes/TertiaryBox";
 import { SpacerRow } from "../../../components/spacer";
-import { getRipperRarity } from "../../../utils/game";
```
n0izn0iz, line 24: maybe find a more explicit name for this file

packages/utils/navigation.ts
```diff
@@ -33,7 +33,7 @@ export type RootStackParamList = {
   RiotGameFight: undefined;
   RiotGameBreeding: undefined;
   RiotGameMemories: undefined;
-  RiotGameMarketplace: undefined;
+  RiotGameMarketplace: { collectionId?: string } | undefined;
```
n0izn0iz, line 36: this seems weird,
I think this might break navigation configuration 
Maybe just use `{ collectionId?: string }`

n0izn0iz (changes requested):

packages/screens/RiotGame/component/RipperSelectorModalV2.tsx
hthieu1110: OK changed it => RipperGridSelectorModal

packages/utils/navigation.ts
hthieu1110: actually without ```| undefined``` it will break ts check in ```navigation.navigate(item.route);``` in ```RiotGameHeader``` file

packages/screens/RiotGame/component/CollectionView.tsx
hthieu1110: Oki. I refactored those components and file ```Marketplace/CollectionScreen```

packages/hooks/riotGame/useGameRewards.ts
```diff
@@ -0,0 +1,62 @@
...
+      );
+      return await distributorQueryClient.userClaimable({
+        addr: selectedWallet.address,
+      });
+    },
+    { staleTime: Infinity }
+  );
+  const claimableAmount = data || 0;
+
+  const claimRewards = async (user: string) => {
```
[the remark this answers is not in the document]
hthieu1110: Done

packages/screens/RiotGame/component/RiotGameHeader.tsx
hthieu1110: We have to support RiotGameMarketplace with and without ```{ collectionId?: string }``` params.
Later on this file we declare a MenuItem like this 
```
  {
    id: "marketplace",
    name: "Marketplace",
    route: "RiotGameMarketplace",
    iconSVG: cartSVG,
  },
```
So I don't file other way than that to support both cases. Might be you have other idea ?

n0izn0iz (dismissed):
Code LGTM

PR #469, opened 2023-03-01
fix(ethereum-backend): doubled nfts

hthieu1110:
Only make 2 queries to filter owner/seller when owner is provided. If not then make just one query.


omniwired:
LGTM :shipit:

omniwired (approved):

n0izn0iz (approved):
LGTM

PR #471, opened 2023-03-02
Allow change pagination parameters Fix #470 [done ✅]

### LINKED ISSUE #470: Marketplace pagination improvements
Allow the pagination to be changed.

Bottom right of this 📺
image

omniwired:
Fix #470 

Needs:
- [x] change the pagination text input into a dropdown
  - [x] Options 50, 100, 200 ?
- [x] review

Looks like this:

image

### DIFF
--- a/packages/components/Pagination.tsx
+++ b/packages/components/Pagination.tsx
@@ -1,33 +1,43 @@
 // libraries
-import React from "react";
+import React, { useRef } from "react";
 import { StyleSheet, TouchableOpacity, View } from "react-native";
 
+import chevronDownSVG from "../../assets/icons/chevron-down.svg";
 import chevronLeftDoubleSVG from "../../assets/icons/chevron-left-double.svg";
 import chevronLeftSVG from "../../assets/icons/chevron-left.svg";
 import chevronRightDoubleSVG from "../../assets/icons/chevron-right-double.svg";
 import chevronRightSVG from "../../assets/icons/chevron-right.svg";
-import { neutral77 } from "../utils/style/colors";
-import { fontSemibold14 } from "../utils/style/fonts";
+import chevronUpSVG from "../../assets/icons/chevron-up.svg";
+import { useDropdowns } from "../context/DropdownsProvider";
+import {
+  neutral33,
+  neutral77,
+  neutralA3,
+  secondaryColor,
+} from "../utils/style/colors";
+import { fontSemibold13, fontSemibold14 } from "../utils/style/fonts";
 import { layout } from "../utils/style/layout";
 import { BrandText } from "./BrandText";
 import { SVG } from "./SVG";
+import { SecondaryBox } from "./boxes/SecondaryBox";
 import { TertiaryBox } from "./boxes/TertiaryBox";
-import { TextInputCustom } from "./inputs/TextInputCustom";
 import { SpacerRow } from "./spacer";
 
-// misc
-
-type PaginationProps = {
+interface PaginationProps {
   currentPage: number;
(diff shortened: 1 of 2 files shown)


packages/components/Pagination.tsx
```diff
@@ -17,17 +17,19 @@ import { SpacerRow } from "./spacer";
 
 // misc
 
-type PaginationProps = {
+interface PaginationProps {
   currentPage: number;
   maxPage: number;
-  itemsPerPage: number;
+  itemsPerPage: string;
```
hthieu1110, line 23: I think we should keep this value type = number

packages/components/Pagination.tsx
```diff
@@ -17,17 +17,19 @@ import { SpacerRow } from "./spacer";
 
 // misc
 
-type PaginationProps = {
+interface PaginationProps {
```
WaDadidou, line 20: Why using `interface` instead of `type` ? (In general)

packages/components/Pagination.tsx
```diff
@@ -95,7 +97,9 @@ export const Pagination = ({
           name="items"
           label=""
           variant="labelOutside"
-          value={itemsPerPage.toString()}
+          onChangeText={(s) => setItemsPerPage(parseInt(s, 10))}
+          onPressEnter={() => setItemsPerPage(parseInt(itemsPerPage, 10))}
+          value={itemsPerPage}
```
WaDadidou, line 102: When the input is empty, the displayed value is "NaN"

packages/components/Pagination.tsx
omniwired: in general:

https://blog.devgenius.io/interfaces-vs-types-in-typescript-7567ffe390cd

packages/components/Pagination.tsx
omniwired: changed to be a dropdown

packages/components/Pagination.tsx
WaDadidou: interface can be extandable, ok thx !

WaDadidou:
I'm sorry, could you add more space between these buttons ?
Because we can see the box's broken corner
image

packages/components/Pagination.tsx
```diff
@@ -83,20 +96,68 @@ export const Pagination = ({
...
-      >
-        <BrandText style={styles.grayText}>Items per page:</BrandText>
-        <TextInputCustom<{ items: number }>
-          width={80}
-          name="items"
-          label=""
-          variant="labelOutside"
-          value={itemsPerPage.toString()}
-        />
+      <View style={styles.section}>
```
WaDadidou, line 99:
```suggestion
      <View style={[
          styles.section,
          { width: edgeSectionsWidth, justifyContent: "flex-end" },
        ]}>
```

It will allow to perfect center the navigation buttons

packages/components/Pagination.tsx
```diff
@@ -75,28 +81,76 @@ export const Pagination = ({
...
+                    flexDirection: "row",
+                    alignItems: "center",
+                    marginBottom: layout.padding_x1_5,
+                  },
+                ]}
+              >
+                <BrandText
+                  style={[fontSemibold13, { marginLeft: 12, color: neutralA3 }]}
+                >
+                  {item.toString(10)}
```
hthieu1110, line 148: `{item}` should do the work

WaDadidou (approved):

zxxma:
LGTM 👍  (ok for bypass branch protections)

PR #482, opened 2023-03-14
Allow Right click and Meta (control,command) open in a new tab [done ✅]

### LINKED ISSUE #478: Can't open NFT page in a new tab
From the community:

_Another suggestion I have is to make it possible to "open in a new tab" the NFTs that we want to have a look at, so we don't have to re-load the entire market every time after having a look at an NFT we're interested in. This really makes sniping and shopping super bothersome._ 

image

omniwired:
Fixes #478

Demo here:

https://user-images.githubusercontent.com/322761/224984270-2a2c7a21-771a-46a9-a699-5dcc07caa308.mp4

https://user-images.githubusercontent.com/322761/225202739-a94c96fb-02e1-4450-a868-658df5ed5c51.mp4

### DIFF
--- a/packages/components/OmniLink.tsx
+++ b/packages/components/OmniLink.tsx
@@ -10,7 +10,7 @@ import {
 } from "react-native";
 
 export interface OmniLinkToType {
-  screen: string;
+  screen: string | never;
   params?: object;
 }
 
@@ -19,10 +19,9 @@ export const OmniLink: React.FC<{
   action?: any | undefined;
   children: ReactNode | undefined;
   style?: StyleProp<ViewStyle | TouchableOpacityProps>;
-
   disabled?: boolean;
-}> = ({ to, action, children, style, disabled, ...rest }) => {
-  // @ts-expect-error
+}> = ({ to, action, children, style, disabled }) => {
+  // @ts-ignore
   const { onPress, ...props } = useLinkProps({ to, action });
 
   const [isHovered, setIsHovered] = React.useState(false);
@@ -34,7 +33,9 @@ export const OmniLink: React.FC<{
     // You can add hover effects using `onMouseEnter` and `onMouseLeave`
     return (
       <View
-        // @ts-expect-error
+        // is required to ignore the following to fix a problem with the linter
+        // and allow to use onClick in this special case
+        // @ts-ignore
         onClick={!disabled ? onPress : null}
         onMouseEnter={() => setIsHovered(true)}
         onMouseLeave={() => setIsHovered(false)}
@@ -46,7 +47,6 @@ export const OmniLink: React.FC<{
           style,
         ]}
         {...props}
-        {...rest}
       >
         {children}
       </View>
@@ -54,7 +54,7 @@ export const OmniLink: React.FC<{
   }
 
   return (
-    <TouchableOpacity style={style} onPress={onPress} {...props} {...rest}>
+    <TouchableOpacity style={style} onPress={onPress} {...props}>
       {children}
     </TouchableOpacity>
   );
--- a/packages/components/OmniLink.tsx
+++ b/packages/components/OmniLink.tsx
@@ -1,0 +1,61 @@
+import { useLinkProps } from "@react-navigation/native";
+import React, { ReactNode } from "react";
+import {
+  Platform,
+  StyleProp,
+  TouchableOpacity,
+  TouchableOpacityProps,
+  View,
+  ViewStyle,
+} from "react-native";
+
+export interface OmniLinkToType {
+  screen: string;
+  params?: object;
+}
+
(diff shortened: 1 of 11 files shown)


packages/components/OmniLink.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+export const OmniLink: React.FC<{
+  to: OmniLinkToType;
+  action?: any | undefined;
+  children: ReactNode | undefined;
+  style?: StyleProp<ViewStyle | TouchableOpacityProps>;
+
+  disabled?: boolean;
+}> = ({ to, action, children, style, disabled, ...rest }) => {
+  // @ts-expect-error
```
hthieu1110, line 25: We should not hide error here

packages/components/OmniLink.tsx
```diff
@@ -0,0 +1,61 @@
...
+}
+
+export const OmniLink: React.FC<{
+  to: OmniLinkToType;
+  action?: any | undefined;
+  children: ReactNode | undefined;
+  style?: StyleProp<ViewStyle | TouchableOpacityProps>;
+
+  disabled?: boolean;
+}> = ({ to, action, children, style, disabled, ...rest }) => {
```
hthieu1110, line 24: Basically ...rest should be [] here so it does not make sense to have it in args

packages/components/OmniLink.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const [isHovered, setIsHovered] = React.useState(false);
+
+  if (Platform.OS === "web") {
+    // It's important to use a `View` or `Text` on web instead of `TouchableX`
+    // Otherwise React Native for Web omits the `onClick` prop that's passed
+    // You'll also need to pass `onPress` as `onClick` to the `View`
+    // You can add hover effects using `onMouseEnter` and `onMouseLeave`
+    return (
+      <View
+        // @ts-expect-error
```
hthieu1110, line 37: should not hide error

packages/components/OmniLink.tsx
omniwired: This is a linter issue it will light up as a problem otherwise. 
It's a false positive.
And ideas how to get around it besides the existing lines?

packages/components/OmniLink.tsx
omniwired: Thanks. I was just thinking how to make it as open as possible.
The solution is to delete rest altogether?

packages/components/nfts/NFTView.tsx
```diff
@@ -216,46 +208,53 @@ export const NFTView: React.FC<{
...
                     }}
                   >
-                    {nft.collectionName}
-                  </BrandText>
+                    <NetworkIcon size={12} networkId={nft.networkId} />
+                    <BrandText
+                      numberOfLines={1}
+                      style={{
+                        fontSize: 12,
+                        marginLeft: 10,
```
hthieu1110, line 250: use layout.padding_x

packages/components/OmniLink.tsx
omniwired: improved with a comment, but this is `required`

packages/components/nfts/NFTView.tsx
omniwired: we don't have a 10, the closest is a 12, I will leave it as is.

hthieu1110: In that case I think we should respect the layout so you can use 8 or 12 instead of 10

hthieu1110 (approved):

packages/components/OmniLink.tsx
omniwired: @sakul-budhathoki , @hthieu1110 recommended me your expertise.

What do you think about this line?

I believe that is a limitation of the ts linter, as it does not understand this specific case.
I don't think that is taking into account the `if` I have in place.

If I truly want to fix this, I would just change it to be a native web (`<a>`) element and not a react native element `<View>`

zxxma:
LGTM 👍

packages/components/OmniLink.tsx
WaDadidou: You can use `CustomPressable`.
Add `onMouseEnter?: () => void;` and `onMouseLeave?: () => void;` in `CustomPressableProps`.
Then you can replace this `View onClick` by `CustomPressable onPress`.

Also, we have this component : `packages/components/Link`.

Do you think you can deal with existing components ?

WaDadidou:
Can you add this behavior to CollectionView ? :)

packages/components/OmniLink.tsx
WaDadidou: After discussion, let's try like this !

WaDadidou (approved):

PR #486, opened 2023-03-23
fix: disabled social feed for now

WaDadidou:


zxxma:
Quickfix before real release.

PR #491, opened 2023-03-24
fix: fix UPP width and disabled feed on UPP

WaDadidou:


sakul-budhathoki (approved):

packages/components/hub/HubIntro.tsx
```diff
@@ -51,7 +51,7 @@ const ConnectedIntro: React.FC = () => {
         size="XL"
       />
 
-      <ProfileButton touchableStyle={{ marginTop: 40 }} />
+      <ProfileButton style={{ marginTop: 40 }} />
```
hthieu1110, line 54: Should we use layout.padding ?

hthieu1110 (approved):
LGTM

PR #507, opened 2023-04-06
fix: add needed index to collection

hthieu1110:
`max_supply` and `paused` are used in collections query without indexes
```
... where c.paused = false and c.max_supply != -1 and (select count from count_by_collection where collection_id = c.id) != c.max_supply
```


omniwired (approved):

n0izn0iz (approved):
LGTM

PR #510, opened 2023-04-07
fix indexer  for social feed contract

hthieu1110:
Due to refactor social feed contract from env to network package, there was a missing part that has not been refactored.


omniwired (approved):

n0izn0iz (approved):
LGTM

PR #513, opened 2023-04-07
feat: responsive ScreenContainer

WaDadidou:
image

## Mobile version added for these components
- ScreenContainer (Don't forget to add a `mobileTitle`)
- SidebarMobile
- NetworkSelector
- HeaderMobile
- TopLogoMobile
- TopMenuMobile
- ConnectWalletMobile

## useIsMobile
- We consider mobile when the window is smaller than 1024px


omniwired:
hey, @WaDadidou thanks for these changes.

I found some issues when going under 1024.
Here is a GIF with the issue:
Kapture 2023-04-10 at 12 23 58

About your RFC about "useIsMobile"

I think we should only care about viewport size and not if the user is Android / iOS.

n0izn0iz (changes requested):

n0izn0iz (approved):
LGTM

PR #529, opened 2023-04-10
fix: better loading for NewsFeed and FeedPostViewScreen

WaDadidou:


omniwired:
@WaDadidou is it MUCH better than before.
Do you know if there is any way to limit these requests? cc @sakul-budhathoki 

Kapture 2023-04-11 at 11 49 40

omniwired (commented):
@WaDadidou is it MUCH better than before.
Do you know if there is any way to limit these requests? cc @sakul-budhathoki 

Kapture 2023-04-11 at 11 49 40

packages/components/socialFeed/NewsFeed/NewsFeed.tsx
```diff
@@ -38,8 +35,10 @@ export const NewsFeed: React.FC<NewsFeedProps> = ({
 }) => {
   const { data, isFetching, refetch, hasNextPage, fetchNextPage, isLoading } =
     useFetchFeed(req);
-  const isLoadingValue = useSharedValue(false);
-  const isGoingUp = useSharedValue(false);
+  const isLoadingValue = useMemo(
```
hthieu1110, line 38: We can assign directly,  I think useMemo does not help much in this case and make the code longer :D

hthieu1110: And my suggestion, the var name can be `isReady` =>
`const isReady = !(isFetching || isLoading)`

packages/screens/FeedPostView/FeedPostViewScreen.tsx
```diff
@@ -60,17 +62,29 @@ export const FeedPostViewScreen: ScreenFC<"FeedPostView"> = ({
...
+    parentId: postResult?.identifier,
+    totalCount: postResult?.sub_post_length,
+  });
   const isNextPageAvailable = useSharedValue(hasNextPage);
 
+  const isLoadingPostValue = useMemo(
+    () => isLoadingPost || isFetchingPost,
+    [isLoadingPost, isFetchingPost]
+  );
+  const isLoadingCommentsValue = useMemo(
```
hthieu1110, line 83: not need useMemo

sakul-budhathoki:
> 
Hi @omniwired  We are getting user info using useNSUserInfo, and those extra requests might be due to that. 
You can check the code below of SocialThreadCard.tsx
```
  const authorNSInfo = useNSUserInfo(localPost.createdBy);
  const userInfo = useNSUserInfo(wallet?.userId);
```

omniwired:
@sakul-budhathoki is there anyway to limit / cache / reduce these kind of calls?

n0izn0iz:
I'm sorry I didn't see this PR, I made the same kind of fix in #537, do we still need this PR, maybe it has other improvements? @WaDadidou

PR #534, opened 2023-04-11
fix: reduce RPCs calls by disabling refetching

hthieu1110:
Disabled refetching for

- [x] useCollections
- [x] useCollectionInfo
- [x] useMintEnded


omniwired (approved):

n0izn0iz (approved):
LGTM

PR #535, opened 2023-04-11
Hot fix hardcode sidebar and apps values

### LINKED ISSUE #532: Sidebar won't load if the backend has problems.
Move the collection of DApps into the front end until the backend is stable or we find another solution

omniwired:
fix #532


WaDadidou (approved):

WaDadidou (dismissed):

n0izn0iz (approved):
LGTM

n0izn0iz (changes requested):
not LGTM, sorry didn't read enough

n0izn0iz:
why are you not combining the hardcoded and fetched results?

omniwired:
@n0izn0iz I don't want the fetch results at all at this time.
I don't want to rely on anything from the backend or airtable.
Actually, I would like to monitor the back end and even disable air table fetch from there as well. What do you think?

omniwired:
> why are you not combining the hardcoded and fetched results?

I've made changes to the code and now is like you suggested.
https://github.com/TERITORI/teritori-dapp/blob/4bc2ccdf453bd47f4dacef36ba900b7712d4eea3/packages/screens/DAppStore/components/DAppStoreData.tsx#L83

packages/screens/DAppStore/query/getFromFile.ts
```diff
@@ -0,0 +1,365 @@
...
+          icon: toripunks,
+          description: "Enter the Bar, play games. punks!",
+          route: "ToriPunks",
+          groupKey: "top-apps",
+          selectedByDefault: false,
+          alwaysOn: false,
+        },
+      },
+    },
+    "coming-soon": {
```
n0izn0iz, line 153: I think this should not be hardcoded, also what happens if one of those is removed from remote source?

n0izn0iz:
merging because scaleway infra is in bad shape

PR #536, opened 2023-04-11
feat: use resizing image proxy for nfts and collections

omniwired:
This is a placeholder with a demo and documentation on how to set up the proxy in our infrastructure.

Guide:

https://www.digitalocean.com/community/tutorials/how-to-serve-next-generation-images-with-imgproxy-using-docker

This is what I did to set it up locally (for testing, just for now).

```
docker pull darthsim/imgproxy:latest

Run it like this

docker run --env IMGPROXY_ENABLE_WEBP_DETECTION=true  -p 8080:8080 -it darthsim/imgproxy

```

I want to set up this properly on the cloud ☁️ that we have.

We can think of some suitable parameters.
The example provided used the value of the image and how big it is actually rendered by the browser to find a good initial step.

### DIFF
--- a/.env
+++ b/.env
@@ -32,3 +32,6 @@ DISPLAYED_NETWORKS_IDS=teritori,teritori-testnet,ethereum,ethereum-goerli,cosmos
 SELECTABLE_NETWORKS_IDS=teritori,teritori-testnet,ethereum,ethereum-goerli,osmosis,osmosis-testnet
 
 PINATA_JWT=your_pinata_admin_jwt_key_here
+
+IMG_PROXY_URL=http://localhost:8080/insecure/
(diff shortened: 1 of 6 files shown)


n0izn0iz (changes requested):
it would be better if we can use an ipfs-aware proxy so we don't depend on a single ipfs gateway anymore

n0izn0iz:
gifs are replaced by a still image :/

otherwise it's pretty fast, nice

n0izn0iz:
found a config for gifs
though now I think the proxy is getting rate limited by nftstorage

n0izn0iz:
with a bit more config, everything looks good

PR #546, opened 2023-04-13
feat: Make the Marketplace Landing page into a grid

omniwired:
Demo:
image


FerrymanGMI:
The R!ot TGEN1 seems missing from the list.
-> Bonus point (sort by total volume).

n0izn0iz:
this must be an infinite scroll, or have another way to scroll all collections IMO

PR #547, opened 2023-04-13
fix(collection screen): better render during load

### LINKED ISSUE #520: Marketplace: collection loading
It shows broken while loading

Demo here:

https://user-images.githubusercontent.com/322761/230864642-d7cbb941-bcfe-4983-a03d-d0acc0992ecf.mp4

omniwired:
The problem: https://github.com/TERITORI/teritori-dapp/issues/520 

The Fix:

https://user-images.githubusercontent.com/322761/231625871-44a2ee47-0b9c-4a25-8e2e-e7422028b129.mp4

### How to test it

- Visit https://deploy-preview-547--testitori.netlify.app/ or your local
- Navigate to a collection in the marketplace
- Check the header issue detailed in the video in this PR
- Review that the issue is not present anymore.
- Optional: compare production and this PR for differences.

### DIFF
--- a/packages/components/collections/CollectionHeader.tsx
+++ b/packages/components/collections/CollectionHeader.tsx
@@ -138,7 +138,7 @@ export const CollectionHeader: React.FC<{
   };
 
   // returns
-  return (
+  return width > 0 ? (
     <View style={{ maxWidth: width, alignSelf: "center" }}>
       <Image
         source={
@@ -262,7 +262,7 @@ export const CollectionHeader: React.FC<{
         />
       </PrimaryBox>
     </View>
-  );
+  ) : null;
 };
 
 const styles = StyleSheet.create({


omniwired:
@WaDadidou any ideas how now to use this value like this
```
minWidth: 720,
```
Maybe using the width of the screen but that might be not fast enough.
Any other ideas, maybe using only CSS?

n0izn0iz:
Screenshot 2023-04-17 at 14 01 35
I get this during load, and it can even stay in this state sometimes

omniwired:
Wow really?
Which browser and resolution? I will have a look @n0izn0iz

omniwired:
> Screenshot 2023-04-17 at 14 01 35 I get this during load, and it can even stay in this state sometimes

### Nice find
thank you so much for your testing.
The old solution was not correct.
#### I've updated the code to fix the root issue:

The hook `useMaxResolution()` comes back with a 0 while loading, this causes the UI to look distorted for a second

WaDadidou (approved):

n0izn0iz (approved):
LGTM

PR #550, opened 2023-04-13
fix #525 Wallet page: Disconnect and rename button doesn't work

### LINKED ISSUE #525: Wallet page: Disconnect and rename button doesn't work
How to reproduce:
Visit:
https://app.teritori.com/wallet-manager

https://user-images.githubusercontent.com/322761/230880616-92c68e7d-6ce7-4c36-b1b7-00990fcdac1c.mp4

omniwired:
## Notes:

- We don't have the notion of rename wallets
- also we don't allow to have more than one keplr wallet. (Keplr has it's own selector)
- I did a quick research and it can't be done.
- Implemented disconnect
- Deleted rename


n0izn0iz:
remove will be added, it's a teaser

you didn't implement disconnect, and it's not possible with keplr (maybe it changed now but it wasn't the case earlier)
we don't want to trick the user into feeling it's wallet is really "disconnected", anyone using it's computer can still make a tx in this state

you can reopen if you feel we still need this in some way

PR #551, opened 2023-04-13
fix(social feed): refresh after posting comment

sakul-budhathoki:
Fixes: https://github.com/orgs/TERITORI/projects/2/views/1?pane=issue&itemId=25401500

Test:

https://user-images.githubusercontent.com/13379067/231695320-dbb4c4c2-72b2-4ba4-9e40-845981ffb3d8.mov


omniwired:
looks, good, can we have a way to maybe scroll into it, or make it fade in?
what do you think @FerrymanGMI ?

PR #559, opened 2023-04-16
fix: toripunks dapp update

omniwired:


n0izn0iz:
why can't they open a PR against their fork?

omniwired:
@n0izn0iz Hi, this is what they did before
https://github.com/TERITORI/teritori-dapp/pull/479
But each of us must request access to their fork if any changes are needed.
Cross-fork PRs how we started this collaboration. I've just moved the code to our fork so we can review it and make the necessary changes.

This was a workaround that needed to happen as the release time was soon. But we can go back to the way things were.

## For this PR specifically, 
I had to diverge from `main` and reconcile the changes.
The reason is: They weren't aware of Teritori's team changes, and the branches became full of conflicts.

I've communicated to diverge from "prod", which is our main branch if they want to do extra changes.

This shouldn't be a problem going forward.

PR #572, opened 2023-04-20
fix: use grpc-web default transport for safari support

### LINKED ISSUE #556: Make gRPC work on safari: make website work on iPhone
More on this:

https://developer.apple.com/forums/thread/694113
https://github.com/gorilla/websocket/issues/731

image

The backend is not working on iPhone

# Notes

This is the problematic one

image

We need a fallback or changes to make it compatible with Iphone.

omniwired:
fix #556 

- I also tried to fix it via golang at gorilla (the lib) level.
- I also tried setting different values in the backend for the gRPC, but this is the only reliable way, I found, to do it.

## How to test

- Open the website in Safari (mobile and desktop).
- The sidebar and everything else should *NOW* work.
- Test other browsers for regressions.

## Important notes
- Needs this line commented out https://github.com/TERITORI/teritori-dapp/blob/8fba61e64cd92e1e7829d6fa4a0fd4d0f61d39b7/packages/context/SidebarProvider.tsx#L73 (addressed at #535 )
- I didn't try to send any tx using keplr.

### DIFF
--- a/packages/utils/backend.ts
+++ b/packages/utils/backend.ts
@@ -26,7 +26,7 @@ export const getMarketplaceClient = (networkId: string | undefined) => {
   }
   if (!marketplaceClients[network.id]) {
     const rpc = new MarketplaceGrpcWebImpl(network.backendEndpoint, {
-      transport: grpc.WebsocketTransport(),
+      // transport: grpc.WebsocketTransport(),
       debug: false,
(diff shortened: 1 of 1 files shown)


omniwired:
@n0izn0iz I will assign this one to you to check the possible implications of this change.

n0izn0iz:
we don't use client-streaming or bi-directional streams so this is ok

PR #584, opened 2023-04-25
feat: "buy tokens" button

omniwired:
The Osmosis website doesn't have a buy tokens button on mobile.

https://user-images.githubusercontent.com/322761/234165433-cac4bee4-b57e-40fc-9d90-29a8391f7c24.mp4

### DIFF
--- a/packages/components/navigation/BuyTokens.tsx
+++ b/packages/components/navigation/BuyTokens.tsx
@@ -1,0 +1,30 @@
+import React from "react";
+import { Linking, TouchableOpacity } from "react-native";
+import { CreditCardIcon } from "react-native-heroicons/outline";
+
+import { primaryColor } from "../../utils/style/colors";
+import { fontBold16 } from "../../utils/style/fonts";
+import { layout } from "../../utils/style/layout";
+import { BrandText } from "../BrandText";
+
+export const BuyTokens: React.FC = () => {
+  return (
+    <TouchableOpacity
+      style={{
+        justifyContent: "center",
+        alignItems: "center",
+        marginBottom: layout.padding_x1,
+        flexDirection: "row",
+      }}
+      onPress={() => {
+        Linking.openURL("https://frontier.osmosis.zone/?from=OSMO&to=TORI");
+      }}
+    >
+      <CreditCardIcon
+        color={primaryColor}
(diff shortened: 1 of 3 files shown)


packages/components/navigation/BuyTokens.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { fontBold16 } from "../../utils/style/fonts";
+import { BrandText } from "../BrandText";
+
+export const BuyTokens: React.FC = () => {
+  return (
+    <TouchableOpacity
+      style={{
+        justifyContent: "center",
+        alignItems: "center",
+        marginBottom: "0.5em",
```
WaDadidou, line 15: You should use the same space value as the other eleemnts in the Sidebar.
Like `padding.layout_x1`

packages/components/navigation/BuyTokens.tsx
```diff
@@ -0,0 +1,26 @@
...
+        flexDirection: "row",
+      }}
+      onPress={() => {
+        Linking.openURL("https://frontier.osmosis.zone/?from=OSMO&to=TORI");
+      }}
+    >
+      <CreditCardIcon color={primaryColor} style={{ marginRight: "0.5em" }} />
```
WaDadidou, line 22: You should use the same space value as the other elements in the Sidebar.
Like padding.layout_x1

omniwired:
> Could you make this design more "unified", regarding the existing buttons in the Sidebar ? Like :
> 
> * Same font (mainly)
> * Same horizontal align
> * Same icon size
> 
> image

Actually I want it to stand up.
Be different

WaDadidou (changes requested):

packages/components/navigation/BuyTokens.tsx
n0izn0iz: furthermost `em` is not supported by native

packages/components/navigation/BuyTokens.tsx
n0izn0iz: no `em` see https://reactnative.dev/docs/height-and-width

n0izn0iz (changes requested):

omniwired:
@WaDadidou I tried setting up the font `fontSemibold12` and is too small. IMO

PR #598, opened 2023-04-27
fix(sidebar): make Social Feed icon appear on top position

omniwired:


packages/context/SidebarProvider.tsx
```diff
@@ -45,22 +45,24 @@ export const SidebarContextProvider: React.FC = ({ children }) => {
   useEffect(() => {
     if (selectedApps.length === 0 && Object.values(availableApps).length > 0) {
       dispatch(
-        setSelectedApps(
-          Object.values(availableApps).flatMap((item) => {
+        setSelectedApps([
+          "top-apps*SEPARATOR*social-feed",
```
n0izn0iz, line 49: why do we need this line?
can't it just respect the definition order?

omniwired: The grouping of each category is causing the order to be mixed up.

PR #599, opened 2023-04-27
feat: wallet connect support

n0izn0iz:


omniwired:
@n0izn0iz I was wondering if this PR is for v1 or v2.
https://docs.walletconnect.com/2.0/advanced/migration-from-v1.x/overview

n0izn0iz:
IIRC keplr does not support v2

Issue #595, opened 2023-04-26
Social Feed: issues when dealing with Video


omniwired:
some videos don't upload properly

This is the video from the user.

https://teritori-dev1.omniwired.com/%D0%A4%D0%B8%D0%BD%D0%B0%D0%BB.mp4

Example:
https://app.teritori.com/feed/post/23b4a17d-0889-4116-be27-773e9d397b42

## Initial thought

The video is over 300 MB, we should not allow big videos or at least show a warning.

WaDadidou:
A PR has been merged reltaed to this issue.
Also, a lot of fixes have been made. I close the issue

PR #602, opened 2023-04-28
fix: Social Feed files

WaDadidou:
- Don't create the post and show an error if pinToIPFS fails
- Fix Audio not found
- Fix `NewsFeedInput` width in `CreateShortPostModal`
- Fix"Videos don't appear after Article creation"


omniwired:
hi @WaDadidou is this one fixed https://github.com/TERITORI/teritori-dapp/issues/595 ? 

Maybe is related to this
> Don't create the post and show an error if pinToIPFS fails

PR #605, opened 2023-05-09
fix: Search button mobile

WaDadidou:


packages/context/SearchBarProvider.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { selectSearchText } from "../store/slices/search";
+import { getMarketplaceClient } from "../utils/backend";
+
+interface DefaultValue {
+  isSearchModalMobileOpen: boolean;
+  setSearchModalMobileOpen: (value: boolean) => void;
+  names?: string[];
+  collections?: Collection[];
+  hasCollections?: boolean;
+  hasNames?: boolean;
```
n0izn0iz, line 16:
```suggestion
  names: string[];
  collections: Collection[];
  hasCollections: boolean;
  hasNames: boolean;
```

we can provide sane default values for these so I recommend to not allow them to be undefined

n0izn0iz (changes requested):

packages/context/SearchBarProvider.tsx
WaDadidou: Fixed

n0izn0iz:
also the search is opened on page load on mobile

n0izn0iz (approved):
LGTM, please fix lint warning

packages/components/Search/SearchBarInput.tsx
```diff
@@ -0,0 +1,57 @@
...
+          {
+            color: "white",
+            flex: 1,
+            marginLeft: 8,
+          },
+          { outlineStyle: "none" } as any,
+        ]}
+        onChangeText={(text) => {
+          dispatch(setSearchText(text));
+          onInteraction && onInteraction();
```
hthieu1110, line 48: Shortcut: onInteraction?.();

hthieu1110 (approved):
LGTM

PR #610, opened 2023-05-16
Simple not enough funds

zerogamma:


omniwired (approved):

packages/screens/DAppStore/apps/toripunks/content/RussianRoulette.tsx
```diff
@@ -177,6 +181,15 @@ export const Russian = () => {
       return setResult(!result);
     }
     if (!bet) return;
+    if (balance.length > 0) {
+      if (
+        balance[0].denom === "utori" &&
```
hthieu1110, line 186: if user does not have utori then we by pass this check ? 
According to this check, we check only if denom = utori, is that expected behaviour ?

omniwired: yes, the toripunks app only cares about TORI

packages/screens/DAppStore/apps/toripunks/content/RussianRoulette.tsx
```diff
@@ -177,6 +181,15 @@ export const Russian = () => {
...
+        parseInt(balance[0].amount, 10) < bet * 1000000
```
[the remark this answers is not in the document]
omniwired: won't fix

hthieu1110 (approved):
LGTM

zxxma:
LGTM

PR #616, opened 2023-05-25
fix: cover image format at article creation

WaDadidou:
At the article creation, the cover image is now at the same format as the articles in the feed.
It prevent the misunderstanding "Why my image is not shown as in the creation (`/feed/new`)" ?


zxxma:
LGTM 👍

omniwired (approved):

PR #626, opened 2023-06-12
minor fixes marketplace v2

omniwired:


omniwired:
@n0izn0iz to fix the collections with denomination issues I propose to write a database migration to just fix the values in the following manner.
TNS:
From denom '' to 'utori'.
Riot tori edition to be "utori".
Diseases from the brain from "atom"(ibc/xxxxxx) to "utori".

n0izn0iz:
The indexer should properly set the denom, is it not possible?

go/pkg/marketplace/marketplace_service.go
```diff
@@ -421,12 +421,13 @@ func (s *MarkteplaceService) NFTs(req *marketplacepb.NFTsRequest, srv marketplac
 		if req.IsListed != false {
 			query.Where("is_listed = ?", true) // Options are ALL or buy now
 		}
+
 		if req.PriceRange != nil {
 			if req.PriceRange.Min != 0 {
-				query.Where("price_amount > ?", req.PriceRange.Min)
+				query.Where("price_amount >= ?", req.PriceRange.Min*1000000) // todo: resolve denom decimals
```
hthieu1110, line 427: Why we don't use denom decimals im this PR ? :o

packages/screens/Marketplace/CollectionScreen.tsx
```diff
@@ -24,6 +27,7 @@ export const CollectionScreen: ScreenFC<"Collection"> = ({ route }) => {
   const { id } = route.params;
   const [selectedTab, setSelectedTab] = useState<TabsListType>("collections");
   const { collectionInfo: info } = useCollectionInfo(id);
+  console.log(info);
```
hthieu1110, line 30: remove if not needed

n0izn0iz:
- we shouldn't need to use float conversions, or at least warn that it's temporary
- using the networks first currency is wrong, you have to get the real denom of "price_amount"
- injecting "decimals" in `networks.json` for one ibc currency is bad, we already have the decimals from the native currency definition, you can do something like `GetNativeCurrency(...).Decimals`

I think the PriceRange filter should already be in atomics and have the denom passed from client

PR #635, opened 2023-06-30
fix: sidebar minimized on load

omniwired:


packages/store/slices/settings.ts
```diff
@@ -20,6 +21,7 @@ const initialState: Settings = {
   isAdenaConnected: false,
   alreadyVisited: false,
   areTestnetsEnabled: false,
+  sideBarExpanded: true,
```
n0izn0iz, line 24: wouldn't this have a bad behavior on mobile? (burger menu is open the first time you start the app)

omniwired: I tried on my phone, but that is not the case.
The sidebar is not expanded on load.
Is following the viewport size in here:

```
  useEffect(() => {
    setIsSidebarExpanded(
      windowWidth >= MOBILE_WIDTH ? settingIsSidebarExpanded : false
    );
  }, [settingIsSidebarExpanded, windowWidth]);
  ```

File: packages/context/SidebarProvider.tsx [hunk 2]
```diff
@@ -35,6 +39,8 @@ const MOBILE_WIDTH = 768;
 
 export const SidebarContextProvider: React.FC = ({ children }) => {
   // The entered isSidebarExpanded
+  const settingIsSidebarExpanded = useSelector(selectSidebarExpanded);
+
   const [isSidebarExpanded, setIsSidebarExpanded] = useState<boolean>(
```
n0izn0iz, line 44: why keep this context state and not use the redux state directly?

packages/components/navigation/BuyTokens.tsx
```diff
@@ -24,7 +26,7 @@ export const BuyTokens: React.FC = () => {
         color={primaryColor}
         style={{ marginRight: layout.padding_x1 }}
```
WaDadidou, line 27:
```suggestion
        style={{ marginRight: flexDirection === "row" ? layout.padding_x1 : 0 }}
```

WaDadidou (changes requested):
On mobile, we have to click twice on the sidebar button to open the sidebar, can you fix it ?

packages/components/navigation/BuyTokens.tsx
WaDadidou, line 27:
```suggestion
        style={{ marginRight: flexDirection === "row" ? layout.padding_x1 : 0 }}
```

WaDadidou (changes requested):

WaDadidou (approved):
LGTM :)

packages/context/SidebarProvider.tsx
```diff
@@ -87,27 +66,14 @@ export const SidebarContextProvider: React.FC = ({ children }) => {
 
     return dynamicAppsSelection;
   }, [availableApps, selectedApps]);
-  const { width: windowWidth } = useWindowDimensions();
-
-  useEffect(() => {
-    setIsSidebarExpanded(
-      windowWidth >= MOBILE_WIDTH ? defaultValue.isSidebarExpanded : false
```
n0izn0iz, line 94: burger menu is now open on first load on small screens

n0izn0iz: I think it make sense to have 2 state, the redux is used on large screen, but on mobile we use a context state, I don't think there is case where we want the menu to be open when we navigate to the site on mobile

omniwired: On load, I could set it to "undefined" by default and then create a condition where it will be closed for mobile breakpoints and open for other screens.
Then once opened or closed, it will follow whatever is on redux.

n0izn0iz: this is nice, but do we really want to keep the menu open on mobile when we close and reopen the app/site ?

omniwired: I understand your point of view, but in order to maintain consistency, I saw we "remember" in both cases. (mobile and desktop).
What do you think?

n0izn0iz (approved):
LGTM

PR #637, opened 2023-07-03
feat: cosmos multisig

WaDadidou:
From https://github.com/TERITORI/teritori-dapp/pull/384
---
___
Buglist 🐛 (some TODO) :

- Looks like "Buy a TNS Account" button doesn't work on Second page after Multisig wallet creation.
==> FIXED
- In "New Delegate Transaction" we should display Validators list in the input, to avoid to find Validators address manually.
==> FIXED
- Plug the Network selector, to allow testnet or mainnet on the feature
==> FIXED
- fix Broadcast status for transactions
==> FIXED
- Allow to use TNS in the multisig creation instead of full adress (easy to add friends then)
==> FIXED
- Once the multisig is created, it should be displayed on Top Right Drop Down menu when we click on Switch Account, and then should allow users to use all dApp features as a 'single wallet', creating tx proposals in Multisig interface.
==> We can select the multisig account, but have to plug miltisig on features
- In "Transactions" screen, we should have displayed all tx, here we received 10 TORI (on mainnet) which are not displayed in this screen.
==> Need to fetch all transactions types, yes

___
quick TODO:

- transaction form, missing labels
==> FIXED
- "Transaction proposals" must become "Transactions" + add "Proposals" tab
==> FIXED
- Prevent aproving transactions if the user is not in the multisig wallet (He musn't have access to the wallet)
==> FIXED
- Enforce to add the multisig creator wallet adress by default at creation (but can be deleted)
==> FIXED

not so quick TODO:
- Allow to give a name to each multisig wallet. Seems to be not possible to hang this is  (Data in backend ? Local ? FaunaDB ?)
- Plugs the behaviors when the user switches to a multisig wallet from TopMenu
- Handle networks
==> FIXED


omniwired:
I leave this just in case this broke more than intended.
https://github.com/TERITORI/teritori-dapp/tree/feat/multisig-wallets-backup

n0izn0iz:
there was a whole go service, you can restore from backup

omniwired:
restored

PR #642, opened 2023-07-10
feat: gno social feeds

hthieu1110:


packages/components/socialFeed/NewsFeed/NewsFeedInput.tsx
```diff
@@ -306,15 +309,55 @@ export const NewsFeedInput = React.forwardRef<
...
+                msg.category.toString(),
+                msg.metadata,
+              ],
+            };
+
+            const txHash = await adenaDoContract(
+              selectedNetworkId,
+              [{ type: "/vm.m_call", value: vmCall }],
+              {
+                gasWanted: 1_000_000,
```
omniwired, line 339: is this OK, hardcoded?

packages/components/socialFeed/NewsFeed/NewsFeedQueries.ts
```diff
@@ -161,17 +161,58 @@ export const createPost = async ({
...
+        msg.category.toString(),
+        msg.metadata,
+      ],
+    };
+
+    const txHash = await adenaDoContract(
+      networkId,
+      [{ type: "/vm.m_call", value: vmCall }],
+      {
+        gasWanted: 2_000_000,
```
omniwired, line 191: ?

packages/components/socialFeed/NewsFeed/NewsFeedInput.tsx
hthieu1110: yes, for this value, default is 1_000_000 so I can remove that

packages/components/socialFeed/NewsFeed/NewsFeedQueries.ts
hthieu1110: yes, 1M does not work so I hardcode to one value that works to make user flow easier, for now I don't see another way to estimate gas efficiently

omniwired (approved):

packages/components/socialFeed/NewsFeed/NewsFeedQueries.ts
n0izn0iz: we can't simulate tx with gno yet

PR #645, opened 2023-07-14
social feed AI

3mp8r3:


omniwired:
it looks quite ready, what's missing?

omniwired:
@FerrymanGMI @zxxma 
At present, there are no user interface elements that inform users of this particular feature.

omniwired (commented):
The code seems fine, but I'm not quite sure how it will insert the responses into the post.

omniwired (approved):

WaDadidou:
I've got the process, but got an error 500 here: https://github.com/TERITORI/teritori-dapp/blob/53efa119243981d453a49eb7918828161ab68b6e/packages/hooks/feed/useBotPost.tsx#L16
With local indexer and TeriotirTestnet

n0izn0iz (changes requested):
this does an ai query on every keystroke??

n0izn0iz:
obsolete

PR #649, opened 2023-08-01
Feat/video player

Saturn-Radius:


packages/screens/VideoPlayer/VideoPlayerHomeContent.tsx
```diff
@@ -0,0 +1,169 @@
...
+    },
+    buttonContainer: {
+      flexDirection: "row",
+      alignItems: "center",
+      paddingLeft: layout.padding_x1,
+      paddingRight: layout.padding_x1_5,
+      paddingVertical: layout.padding_x1,
+      backgroundColor: "#2B2B33",
+      borderRadius: layout.padding_x4,
+      gap: layout.padding_x1_5,
```
WaDadidou, line 102:
```suggestion
      height: 32,
      backgroundColor: neutral30,
      borderRadius: 99,
      gap: layout.padding_x1,
```
(We want a rounded button, so the `borderRadius` can be very height to keep a perfectly rounded button without depending on its size IMO)

packages/components/videoPlayer/UploadVideoModal.tsx
```diff
@@ -0,0 +1,507 @@
...
+    },
+    buttonText: StyleSheet.flatten([
+      fontSemibold14,
+      {
+        color: primaryColor,
+        paddingHorizontal: layout.padding_x1_5,
+        paddingVertical: layout.padding_x1,
+        marginBottom: layout.padding_x2_5,
+        borderRadius: layout.padding_x1,
+        backgroundColor: "#2B2B33",
```
WaDadidou, line 156:
```suggestion
        backgroundColor: neutral30,
```

WaDadidou (changes requested):
We have many differents ways to do things in front in the project, sometimes weird, sorry for that.

- Use `TouchableOpacity` instead of `Pressable`
- Use `layout.padding_xxxx` only for margins and paddings (and gap maybe)
- Use variables `neutralxxx` etc for colors
- Don't hesitate to set static heights, with, spaces for small elements
- Use the current "coding style". It's boring, but it prevent to have too much difference in our front code

WaDadidou:
The video upload seems to not working

PR #650, opened 2023-08-03
fix: correctly use user-provided pinata key + extract upload utils

WaDadidou:
I made this refacto in https://github.com/TERITORI/teritori-dapp/pull/646 and I will need for upcoming WIP features.
Especially this function (Get key from bakcend + upload file) : https://github.com/TERITORI/teritori-dapp/blob/e0e312eff68d0113f1273522980b19533c9c3475/packages/utils/ipfs.ts#L64

Also, the users can now create a key from https://app.pinata.cloud, then provide the JWT directly on the `/settings` page.
(No backend call if you provide it)
image


zxxma:
LGTM

PR #657, opened 2023-08-08
Feat: DAO NFT based

WaDadidou:


go/pkg/marketplace/marketplace_service.go
```diff
@@ -1056,3 +1056,37 @@ func (s *MarkteplaceService) SearchCollections(ctx context.Context, req *marketp
...
+	collectionId := req.CollectionId
+
+	type Owners struct {
+		Owner string
+	}
+	var owners []Owners
+
+	_ = s.conf.IndexerDB.Raw(`
+        select distinct owner from (
+            select nfts.id, max(a.time) atime, COALESCE(m.buyer_id, t.buyer_id, tn.receiver) as owner
```
n0izn0iz, line 1071: why aggregate instead of using `owner_id` field?

Makefile
```diff
@@ -16,6 +16,11 @@ RIOTER_FOOTER_PACKAGE=rioter-footer-nft
 VAULT_REPO=teritori-vault
 VAULT_PACKAGE=teritori-nft-vault
 
+DAO_CONTRACTS_REPO=dao-contracts
```
n0izn0iz, line 19: this should not use this repo but instead use upstream DA0-DA0

packages/screens/Organizations/OrganizationDeployerScreen.tsx
```diff
@@ -1,49 +1,62 @@
...
   "Review information",
   "Launch organization",
 ];
 
 export const LAUNCHING_PROCESS_STEPS: LaunchingProcessStepType[] = [
+  { title: "Book name", completeText: "Transaction finalized" },
   { title: "Create organization", completeText: "Transaction finalized" },
+  {
+    title: "Transfer name to organization",
+    completeText: "Transaction finalized",
```
n0izn0iz, line 53: this should do a single transaction in teritori case and keep a single step

n0izn0iz (changes requested):
did not read all yet

Makefile
WaDadidou: You right, fixed !

packages/screens/Organizations/OrganizationDeployerScreen.tsx
WaDadidou: We can only do 2 TXs minimum for now -> need contract revision

PR #665, opened 2023-08-09
fix(gno): feed post details usernames and back button

hthieu1110:


packages/networks/index.ts
```diff
@@ -217,6 +217,8 @@ export const getUserId = (
   const network = getNetwork(networkId);
   if (network?.kind === NetworkKind.Gno && address.startsWith("gno.land/")) {
     address = address.substring("gno.land/".length).replaceAll("/", "-");
+  } else if (address.includes("-")) {
+    address = address.split("-")[1];
```
n0izn0iz, line 221: this is wrong, getUserId is wrongly called if you need this I think

n0izn0iz (changes requested):

PR #666, opened 2023-08-09
fix: handle ns more properly

hthieu1110:


packages/screens/FeedPostView/FeedPostViewScreen.tsx
```diff
@@ -70,11 +70,11 @@ export const FeedPostViewScreen: ScreenFC<"FeedPostView"> = ({
     id,
     selectedNetworkId
   );
-  const authorNSInfo = useNSUserInfo(
-    getUserId(selectedNetworkId, postResult?.post_by)
-  );
+  const userId = getUserId(selectedNetworkId, postResult?.post_by);
```
n0izn0iz, line 73:
```suggestion
  const authorUserId = getUserId(selectedNetworkId, postResult?.post_by);
```

packages/screens/FeedPostView/FeedPostViewScreen.tsx
```diff
@@ -70,11 +70,11 @@ export const FeedPostViewScreen: ScreenFC<"FeedPostView"> = ({
...
+  const authorNSInfo = useNSUserInfo(userId);
 
   const [, userAddress] = parseUserId(postResult?.post_by);
```
n0izn0iz, line 76:
```suggestion
  const authorAddress = postResult?.post_by;
```

n0izn0iz (changes requested):

api/feed/v1/feed.proto
```diff
@@ -27,9 +27,9 @@ message Post {
   string metadata = 4;
   string parent_post_identifier = 5;
   uint32 sub_post_length = 6;
-  string created_by = 7;
   int64 created_at = 8;
   int64 tip_amount = 10;
+  string author_id = 11;
```
n0izn0iz, line 32: keep the same field number to prevent backend downtime

n0izn0iz (changes requested):
keep the same field number

PR #669, opened 2023-08-15
fix: Enhance SelectInput and DAOSelector

WaDadidou:
- Use SelectInput in `DAOSelector`.
- Enhance `SelectInput` and add the possibility to enters a value to search in the `SelectInput` list (Not effective for `DAOSelector`, the image is just an example)

image

image

Before, we used `@react-native-picker/picker`
image

See https://github.com/TERITORI/teritori-dapp/blob/99619c93a12231e8fe551dcfa8d8cb9f84be030f/packages/components/dao/DAOSelector.tsx#L18
And https://github.com/TERITORI/teritori-dapp/blob/99619c93a12231e8fe551dcfa8d8cb9f84be030f/packages/components/inputs/SelectInput.tsx#L52

### DIFF
--- a/packages/components/inputs/SelectInput.tsx
+++ b/packages/components/inputs/SelectInput.tsx
@@ -1,13 +1,14 @@
-import React, { ReactElement, useState } from "react";
+import React, { ReactElement, useEffect, useMemo, useState } from "react";
 import {
   View,
   ScrollView,
   StyleSheet,
   StyleProp,
   ViewStyle,
+  ActivityIndicator,
 } from "react-native";
 
-import { Label } from "./TextInputCustom";
+import { Label, TextInputCustom } from "./TextInputCustom";
 import chevronDownSVG from "../../../assets/icons/chevron-down.svg";
 import chevronUpSVG from "../../../assets/icons/chevron-up.svg";
 import lockSVG from "../../../assets/icons/lock.svg";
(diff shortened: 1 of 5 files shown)


omniwired (dismissed):

packages/hooks/useNSPrimaryAlias.ts
```diff
@@ -87,3 +92,43 @@ export const useNSPrimaryAlias = (userId: string | undefined) => {
   );
   return { primaryAlias: data, ...other };
 };
+
+export const useNSPrimaryAliases = (userIds: string[] | undefined) => {
+  const { data, ...other } = useQuery(
+    nsPrimaryAliasQueryKey(userIds?.map((id) => id.slice(-3)).toString()),
```
n0izn0iz, line 98: this breaks caching, please use other hook deeper in the component tree

packages/components/inputs/SelectInput.tsx
```diff
@@ -224,9 +290,7 @@ const styles = StyleSheet.create({
...
     padding: layout.padding_x1,
   },
-  // dropdownMenuRow: {
-  //   backgroundColor: neutral00,
-  //   borderRadius: 6,
-  //   padding: layout.padding_x1,
-  // },
+  inputIconContainer: {
+    marginLeft: layout.padding_x1_5,
+  },
```
n0izn0iz, line 295: please don't use `stylesheet.Create` or add new fields in `stylesheet.Create`, this creates potential  undetectable deadcode and has no performance impact contrary to popular belief

if you want to define constant styles to prevent re-renders on memoized components, use something like
```ts
const inputIconContainerStyle: ViewStyle = {
  marginLeft: layout.padding_x1_5,
}
```

packages/screens/TeritoriNameService/TNSManageScreen.tsx
```diff
@@ -71,7 +71,6 @@ export const TNSManageScreen: React.FC<TNSManageScreenProps> = ({
             </BrandText>
 
             <DAOSelector
-              value={selectedDAOId}
```
n0izn0iz, line 74: are you sure about this change? uncontrolled inputs can be a pain

n0izn0iz (changes requested):
did not read everything yet

PR #672, opened 2023-08-18
 fix: Enhance SelectInput and DAOSelector

WaDadidou:
- Use SelectInput in `DAOSelector`.
- Enhance `SelectInput` and add the possibility to enters a value to search in the `SelectInput` list (Not effective for `DAOSelector`, the image is just an example)

image

image

Before, we used `@react-native-picker/picker`
image

See https://github.com/TERITORI/teritori-dapp/blob/99619c93a12231e8fe551dcfa8d8cb9f84be030f/packages/components/dao/DAOSelector.tsx#L18
And https://github.com/TERITORI/teritori-dapp/blob/99619c93a12231e8fe551dcfa8d8cb9f84be030f/packages/components/inputs/SelectInput.tsx#L52

### DIFF
--- a/packages/components/dao/DAOSelector.tsx
+++ b/packages/components/dao/DAOSelector.tsx
@@ -1,53 +1,62 @@
-import { Picker } from "@react-native-picker/picker";
-import { ComponentProps } from "react";
+import React, { useMemo, useState } from "react";
+import { StyleProp, ViewStyle } from "react-native";
 
 import { useDAOs } from "../../hooks/dao/useDAOs";
-import { useNSPrimaryAlias } from "../../hooks/useNSPrimaryAlias";
+import { useNSPrimaryAliases } from "../../hooks/useNSPrimaryAlias";
 import { parseUserId } from "../../networks";
-import { neutral77 } from "../../utils/style/colors";
-import { fontSemibold12 } from "../../utils/style/fonts";
+import { tinyAddress } from "../../utils/text";
+import { SelectInput, SelectInputData } from "../inputs/SelectInput";
+
+const ownWalletToSelect: SelectInputData = {
+  label: "Use my wallet",
+  value: "",
+};
 
 export const DAOSelector: React.FC<{
   userId: string | undefined;
-  value: string;
-  onSelect: (address: string) => void;
-  style?: ComponentProps<typeof Picker>["style"];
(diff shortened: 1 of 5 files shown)


omniwired (dismissed):

omniwired (dismissed):

packages/components/dao/DAOSelector.tsx
```diff
@@ -1,53 +1,64 @@
...
-    >
-      <Picker.Item label="Use my wallet" value="" />
-      {(daos || []).map((dao) => {
-        return <DAOPickerItem key={dao.id} daoId={dao.id} />;
-      })}
-    </Picker>
+  // We get the aliases for all these daos and provide them in the SelectInput. We display aliases or addresses and use userId as value
+  const { primaryAliases: daosAliases, isLoading } = useNSPrimaryAliases(
+    daos?.map((d) => d.id)
+  );
```
n0izn0iz, line 30: this breaks caching

n0izn0iz (changes requested):

packages/components/inputs/SelectInput.tsx
```diff
@@ -37,34 +45,84 @@ type Props = {
   data: SelectInputData[];
   placeHolder?: string;
   selectedData: SelectInputData;
-  setData: (data: SelectInputData) => void;
+  selectData: (data: SelectInputData) => void;
```
n0izn0iz, line 48:
```suggestion
  onSelect: (data: SelectInputData<T>) => void;
```

n0izn0iz: also it would better to use a generic for the data type https://www.typescriptlang.org/docs/handbook/2/generics.html

n0izn0iz, line 48: name is weird
```suggestion
  onSelect: (data: SelectInputData) => void;
```
also would be better to use a generic type for data value https://www.typescriptlang.org/docs/handbook/2/generics.html
```suggestion
  selectData: (data: SelectInputData<T>) => void;
```

n0izn0iz (changes requested):

packages/components/inputs/SelectInput.tsx
WaDadidou: type simplified (only string)

PR #581, opened 2023-04-24
WIP: Desktop & App build

sakul-budhathoki:

PR #673, opened 2023-08-21
chore: update quick-start instructions

n0izn0iz:


ValentinPy1 (commented):
Maybe putting a warning that desktop is in work in progress

n0izn0iz:
removed the desktop instructions since the new desktop version does not uses expo-electron I believe 
https://github.com/TERITORI/teritori-dapp/pull/581

PR #686, opened 2023-09-01
refactor(styles): Layout -> padding into spacing

omniwired:


hthieu1110 (approved):
LGTM, this should be merged first so we can fixed other PRs with this change :)

PR #690, opened 2023-09-05
fix: social feed audio size fixed

WaDadidou:
## Before
image

## After
image

## Mobile (after)
image


packages/components/FilePreview/AudioView.tsx
```diff
@@ -192,8 +188,8 @@ export const AudioView: React.FC<{
               }}
               resizeMode="cover"
               style={{
-                height: THUMBNAIL_WIDTH,
-                width: THUMBNAIL_WIDTH,
+                height: 140,
+                width: 140,
```
n0izn0iz, line 192: can you add back the constant pls? named variables are better than hardcoded numbers, you can add it in this file directly

n0izn0iz (changes requested):

PR #697, opened 2023-09-11
fix: refacto dropdowns

clegirar:
upload_44da99a55aa0aec26cc81cf2abf9e0f9
upload_d44697e79c48bc1c39fa4983dfc2ff23
Screenshot 2023-09-11 at 17 03 15

This `tree dots` components were not consistant, especially when we clicked outside of the dropdown, all dropdowns had not same behavior

So i gave it back consistant, for the first one (first screen) was good and works well !
For the second one (second screen), you probably just forgot to put the ref in the parent view of the dropdown
For the last one (last image), it was just not implemented :)

Signed-off-by: clegirar <clemntgirard@gmail.com>


packages/components/Menu/Menu.web.tsx
WaDadidou, line 1: Please, remove `Menu` folder and just let a `Menu.tsx` since we have no interest to separate this component into a web version and a RN version, since we have only RN components in the two versions.

packages/components/Menu/Menu.tsx
```diff
@@ -22,51 +23,56 @@ export const Menu: React.FC<MenuProps> = ({
...
+                key={item.label}
+                onPress={() => {
+                  closeOpenedDropdown();
+                  item.onPress();
+                }}
+                activeOpacity={0.7}
                 style={[
-                  {
-                    fontSize: 13,
+                  { paddingVertical: 12, width: "100%" },
```
WaDadidou, line 55:
```suggestion
                  { paddingVertical: layout.spacing_x1_5, width: "100%" },
```
We use variables for paddings and margins to keep consistency (Don't forget to import)

packages/components/Menu/Menu.tsx
```diff
@@ -22,51 +23,56 @@ export const Menu: React.FC<MenuProps> = ({
...
-            >
-              <BrandText
+      {isDropdownOpen(dropdownRef) && (
+        <View ref={dropdownRef}>
+          <PrimaryBox
+            width={width}
+            style={{ position: "absolute", right: 0, bottom: -20 }}
+            mainContainerStyle={{
+              position: "absolute",
+              paddingHorizontal: 12,
```
WaDadidou, line 42:
```suggestion
              paddingHorizontal: layout.spacing_x1_5,
```
We use variables for paddings and margins to keep consistency (Don't forget to import)

packages/components/Menu/Menu.tsx
```diff
@@ -22,51 +23,56 @@ export const Menu: React.FC<MenuProps> = ({
...
-                {item.label}
-              </BrandText>
-            </TouchableOpacity>
-          ))}
-        </PrimaryBox>
+                <BrandText
+                  style={[
+                    {
+                      fontSize: 13,
+                    },
```
WaDadidou, line 66:
```suggestion
                   fontSemiBold13,
```
We use variables for fonts to keep consistency (Don't forget to import)

packages/components/Menu/Menu.web.tsx
clegirar: Do you know why at the beginning we have a web version ? Maybe it exists a reason ?
But i agree with you, i'll test the `Menu.tsx` on both platform to correctly check the behavior 👍

packages/components/Menu/Menu.tsx
clegirar: I didn't touch to the style yet but okay i can fix that 👍

packages/components/Menu/Menu.web.tsx
WaDadidou: I don't really know, I just see we had a `div` previously

WaDadidou (dismissed):

packages/components/Menu.tsx
```diff
@@ -0,0 +1,76 @@
...
+  items,
+  component,
+  width = DEFAULT_WIDTH,
+}) => {
+  const { onPressDropdownButton, isDropdownOpen, closeOpenedDropdown } =
+    useDropdowns();
+  const dropdownRef = useRef<View>(null);
+
+  return (
+    <View style={{ position: "relative" }}>
```
hthieu1110, line 33: Seems that by default all the View has position relative

hthieu1110 (approved):
LGTM

PR #699, opened 2023-09-12
chore: add ts-unused-exports library

clegirar:
I added ts-unused-exports library to check if we have some unused exports and so some useless components, dead code etc.

Signed-off-by: clegirar <clemntgirard@gmail.com>

### DIFF
--- a/.github/workflows/js.yml
+++ b/.github/workflows/js.yml
@@ -31,3 +31,6 @@ jobs:
 
       - name: Check that there is no diff
         run: git diff --exit-code
+
+      - name: Check typescript unused exports
+        run: make unused-exports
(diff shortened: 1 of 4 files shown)


clegirar:
We have a lot of unused exports ! So how do you want to proceed ? Delete all unused exports ? Comment it ? Or we have to check case by case ?

n0izn0iz:
let's fix everything

clegirar:
Okay 👍 
I just put a TODO comment before the makefile rule to unexclude almost all paths (i specify them in the comment), and i'll fix that in an other PR to keep the implementation and all fixes splited 👍

PR #702, opened 2023-09-13
feat: gno projects_manager

hthieu1110:


zxxma:
🗒️ Note:
**Updating the plan informations about this PR:**

We basically want to update the plan, from **Gno Grant Manager** into **Onchain Project Manager**, which will allow to manage a project in a decentralized way.
By changing both the naming and the vision, we'll buid a more modular feature that can fit:
- any internal or external project management
- for both DAOs, companies, agencies, etc.

This modular vision could be interacting in the future with the decentralized service marketplace, as a module for fair & transparent management of contracts between buyers/sellers, but also for community to follow up projects development/management in a public way.

Currently @ Gnolang Retreat, I'll try to report all our week works, especially the ones related to WorxDAO & SubDAOs workswhops, that could feed this feature.

PR #708, opened 2023-09-25
feat: add ui to burn toris

n0izn0iz:


packages/screens/CoreDAO/CoreDAOScreen.tsx
```diff
@@ -116,7 +98,7 @@ const DAOManager: React.FC = () => {
           };
 
           await makeProposal(selectedWallet?.address, {
-            title: "Burn a TORI",
+            title: "Propose to burn a TORI",
```
omniwired, line 101: how about also allowing normal non DAO users?

n0izn0iz:
obsolete

PR #709, opened 2023-09-25
fix(TNSHomeScreen): Fix disabled condition for Register and Manage TNS

omniwired:
- Fix the disabled condition for the Register FlowCard to only disable it when either Keplr or Leap is not connected.
- Fix the disabled condition for the Manage FlowCard to disable it when either Keplr or Leap is not connected, or when there are no tokens available.


packages/screens/TeritoriNameService/TNSHomeScreen.tsx
```diff
@@ -141,7 +142,11 @@ export const TNSHomeScreen: ScreenFC<"TNSHome"> = ({ route }) => {
   return (
     <ScreenContainer
       noMargin={width <= 1600}
-      headerChildren={<BrandText>Name Service</BrandText>}
+      headerChildren={
+        <BrandText style={fontSemibold20}>Name Service</BrandText>
+      }
+      responsive
+      onBackPress={() => navigation.navigate("TNSHome")}
```
WaDadidou, line 149: We already are on TNSHome.
Maybe should we navigate to "Home" on root feature's screens ?

WaDadidou (changes requested):

PR #714, opened 2023-10-10
feat: theming foundations

clegirar:
Add light theme logic and a little example on how use it 👍

### DIFF
--- a/packages/components/BrandText/BrandTextBase.tsx
+++ b/packages/components/BrandText/BrandTextBase.tsx
@@ -1,11 +1,15 @@
 import React from "react";
 import { Text, TextProps, StyleSheet } from "react-native";
 import TextTicker from "react-native-text-ticker";
+import { useSelector } from "react-redux";
+
+import { selectIsLightTheme } from "../../store/slices/settings";
 
 export const BrandTextBase: React.FC<TextProps & { isTicker?: boolean }> = (
   props
 ) => {
   const { style, isTicker, ...otherProps } = props;
+  const styles = useStyles();
 
   if (isTicker)
     return (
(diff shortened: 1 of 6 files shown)


omniwired (approved):

packages/components/BrandText/BrandTextBase.tsx
```diff
@@ -26,12 +30,15 @@ export const BrandTextBase: React.FC<TextProps & { isTicker?: boolean }> = (
...
-  base: {
-    color: "white",
-    fontSize: 20,
-    fontWeight: "600",
-  },
-});
+const useStyles = () => {
+  const isLightTheme = useSelector(selectIsLightTheme);
+  // eslint-disable-next-line no-restricted-syntax
+  const styles = StyleSheet.create({
```
n0izn0iz, line 36: remove the stylesheet.create pls

n0izn0iz (changes requested):

packages/components/BrandText/BrandTextBase.tsx
clegirar: Why ?
There is a lot of `StyleSheet.create` all over the codebase

clegirar: (I changed it but i would have a reason)

n0izn0iz: if you hover on the error it will explain the rationale

n0izn0iz: main thing is that it breaks type safety

n0izn0iz: one dev forced it into the repo then cargo cult made it appear everywhere, but I was against it since the start, we added the linter check recently because we decided to stop doing that

clegirar: Okay okay 👍 for type reason so

packages/components/BrandText/BrandTextBase.tsx
```diff
@@ -26,12 +30,17 @@ export const BrandTextBase: React.FC<TextProps & { isTicker?: boolean }> = (
...
-const styles = StyleSheet.create({
-  base: {
-    color: "white",
-    fontSize: 20,
-    fontWeight: "600",
-  },
-});
+const useStyles = () => {
+  const isLightTheme = useSelector(selectIsLightTheme);
+  // eslint-disable-next-line no-restricted-syntax
```
WaDadidou, line 35: This comment is used to prevent an error due to StyleSheet.create, since we didn't yet  replaced all StyleSheet.create. So you can remove it
```suggestion
```

WaDadidou (changes requested):

packages/components/BrandText/BrandTextBase.tsx
clegirar: Yeah totally true 👍

PR #716, opened 2023-10-13
feat(sidebar): make "buy tokens" redirect to swap screen and enable "+" button

omniwired:
This is a suggestion, there are 2 changes.
1- Since Osmosis got rid of "frontier", we got a much worse user experience to onboard users. Specially buying their first tori.
This proposed change is leveraging the existing Swap feature, so people can use our UI to swap in osmosis.
2- when the users click on the + on the sidebar it will go to dapp store instead of showing coming soon

### DIFF
--- a/packages/components/navigation/BuyTokens.tsx
+++ b/packages/components/navigation/BuyTokens.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { Linking, TextStyle, TouchableOpacity } from "react-native";
+import { TextStyle } from "react-native";
 import { CreditCardIcon } from "react-native-heroicons/outline";
 
 import { primaryColor } from "../../utils/style/colors";
--- a/packages/components/navigation/BuyTokens.tsx
+++ b/packages/components/navigation/BuyTokens.tsx
@@ -5,28 +5,29 @@ import { CreditCardIcon } from "react-native-heroicons/outline";
 import { primaryColor } from "../../utils/style/colors";
 import { layout } from "../../utils/style/layout";
 import { BrandText } from "../BrandText";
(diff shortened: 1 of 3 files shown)


WaDadidou:
About the "+" in the Sidebar, it's a good move, but maybe we could'nt highlight the "+" :face_with_monocle:
image

WaDadidou:
I agree with the Swap redirection, since we redirected to osmosis frontier "from OSMO to TORI"

omniwired:
I thought about it. And I must say that at first it wasn't on purpose.
But I feel it works.
Because the + is the App Store after all!

zxxma:
Agree w/ you both guys 👍

packages/components/navigation/BuyTokens.tsx
```diff
@@ -1,32 +1,33 @@
...
-        Linking.openURL("https://frontier.osmosis.zone/?from=OSMO&to=TORI");
-      }}
     >
       <CreditCardIcon
         color={primaryColor}
         style={{ marginRight: flexDirection === "row" ? layout.spacing_x1 : 0 }}
       />
-      <BrandText style={textStyle}>Buy Tokens</BrandText>
-    </TouchableOpacity>
+      <BrandText style={textStyle}>Buy TORI</BrandText>
```
hthieu1110, line 30: is this alway TORI ? can we get token name from network.currencies[0].displayName or sth like that ?

omniwired: Yes, always tori.

WaDadidou (approved):
Ok LGTM !

hthieu1110 (approved):

PR #718, opened 2023-10-17
feat/notifications: Add NotificationDrawer component

omniwired:


api/notification/v1/notification.proto
```diff
@@ -0,0 +1,24 @@
...
+
+package notification.v1;
+option go_package = "./notificationpb";
+
+service NotificationService {
+  rpc Notifications(NotificationRequest) returns (NotificationResponse);
+}
+
+message NotificationRequest {
+  string network_id = 1;
```
n0izn0iz, line 11: you can get network id from user id

n0izn0iz:
There's one thing to think about, the indexerdb was meant to be a pure aggregation of the chain events, we already introduced some unpure state with p2e but it has been a pain ever since, we should avoid that, so consider using a second db for stateful stuff

go/pkg/notification/notification_service.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+}
+
+func (s *NotificationService) Notifications(ctx context.Context, req *notificationpb.NotificationsRequest) (*notificationpb.NotificationsResponse, error) {
+	query := s.conf.PersistentDB
+	userId := req.GetUserId()
+	if userId == "" {
+		return nil, errors.Wrap(errors.New("need a user id tori-{wallet_address}"), "need a wallet address")
+	}
+	_, updateerr := s.UpdateNotifications(ctx, req)
```
omniwired, line 37: checking for new notifications on every interaction at the moment.

go/pkg/notification/notification_service.go
```diff
@@ -0,0 +1,167 @@
...
+
+		notification := indexerdb.Notification{
+			UserId:    networks.UserID(userId),
+			TriggerBy: d.AuthorId,
+			Body:      "comment-" + d.ParentPostIdentifier + "-" + d.Identifier,
+			Action:    d.ParentPostIdentifier,
+			Category:  "comment",
+			CreatedAt: d.CreatedAt,
+		}
+		s.conf.PersistentDB.Create(&notification)
```
omniwired, line 142: dedup happens in the db itself via a unique index

omniwired:
@n0izn0iz what do you think about adding a network prefix in here https://github.com/TERITORI/teritori-dapp/blob/005c83df235a8ce00d5a16275b3b665bb9cf4d07/go/internal/indexerhandler/feed.go#L183

We could have a formatted like this one
```
func (n *NetworkBase) PostID(identifier string) PostID {
	return PostID(fmt.Sprintf("%s-%s", n.IDPrefix, identifier))
}
```

packages/components/TopMenu/NotificationDrawer.tsx
```diff
@@ -0,0 +1,370 @@
...
+            size="S"
+            style={{ width: "100%" }}
+          />
+        </View>
+
+        <View style={{ flex: 4, marginHorizontal: layout.spacing_x1 }}>
+          <OmniLink
+            to={{
+              screen: "FeedPostView",
+              params: { id: item.action },
```
WaDadidou, line 239: Whythis property is named "action" ? What was the original purpose ?

packages/components/TopMenu/NotificationDrawer.tsx
```diff
@@ -0,0 +1,370 @@
...
+                    fontSemibold8,
+                    {
+                      position: "absolute",
+                      right: 2.3,
+                    },
+                  ]}
+                >
+                  {notifications.length > 9 ? "♾️" : notifications.length}
+                </BrandText>
+              </View>
```
WaDadidou, line 101: I purpose something like this

```suggestion
                            <View
                style={{
                  position: "absolute",
                  top: -2,
                  right: -2,
                  alignItems: "center",
                  justifyContent: "center",
                  borderRadius: 9999,
                  height: 14,
                  paddingHorizontal: layout.spacing_x0_75,
                  backgroundColor: primaryColor,
                }}
              >
                <BrandText
                  style={[
                    fontSemibold10,
                    {
                    },
                  ]}
                >
                  {notifications.length > 99 ? "+99" : notifications.length}
                </BrandText>
              </View>
```

packages/components/TopMenu/NotificationDrawer.tsx
```diff
@@ -0,0 +1,370 @@
...
+const useBuildBodyText = (item: Notification) => {
+  const userInfo = useNSUserInfo(item.triggerBy);
+  const name =
+    userInfo?.metadata?.tokenId || tinyAddress(item.triggerBy, 30) || "";
+
+  if (item.category === "reaction") {
+    return `${name} ${item.body} your post.`;
+  }
+  if (item.category === "comment") {
+    return `${name} commented your post.`;
```
WaDadidou, line 338: I purpose somthing like this (address used in tinyAddress instead of userId + linebreak)
```suggestion
    const [, address] = parseUserId(item.triggerBy)
  const name =
    userInfo?.metadata?.tokenId || `${tinyAddress(address, 24)}` || "";

  if (item.category === "reaction") {
    return `${name}\n${item.body} your post.`;
  }
  if (item.category === "comment") {
    return `${name}\ncommented your post.`;
```

WaDadidou (changes requested):

packages/components/TopMenu/NotificationDrawer.tsx
```diff
@@ -0,0 +1,370 @@
...
+        <BrandText style={fontSemibold14}>Clear All</BrandText>
+      </TouchableOpacity>
+    </View>
+  );
+};
+
+const useBuildBodyText = (item: Notification) => {
+  const userInfo = useNSUserInfo(item.triggerBy);
+  const name =
+    userInfo?.metadata?.tokenId || tinyAddress(item.triggerBy, 30) || "";
```
WaDadidou, line 332:
```suggestion
    const [, address] = parseUserId(item.triggerBy)
  const name =
    userInfo?.metadata?.tokenId || `${tinyAddress(address, 24)}` || "";
```

WaDadidou (changes requested):

n0izn0iz:
> @n0izn0iz what do you think about adding a network prefix in here
> 
> https://github.com/TERITORI/teritori-dapp/blob/005c83df235a8ce00d5a16275b3b665bb9cf4d07/go/internal/indexerhandler/feed.go#L183
> 
> We could have a formatted like this one
> 
> ```
> func (n *NetworkBase) PostID(identifier string) PostID {
> 	return PostID(fmt.Sprintf("%s-%s", n.IDPrefix, identifier))
> }
> ```

Sure

packages/components/TopMenu/NotificationDrawer.tsx
omniwired: can you provide a screenshot side by side? thanks in advance

omniwired:
@n0izn0iz https://github.com/TERITORI/teritori-dapp/pull/718/commits/4e73872fe4caab61eb148da80de833dbc64a9319 this is the change to the posts table to make the "ids" to be network aware.

PR #727, opened 2023-11-02
Feat/follow user

omniwired:
placeholder


omniwired:
@n0izn0iz 
doing some necromancy here, lets not forget about this one.

PR #740, opened 2023-11-14
feat(messenger): redux and weshnet connection

sakul-budhathoki:


packages/weshnet/message/subscriber.ts
```diff
@@ -0,0 +1,88 @@
...
+import {
+  GroupMessageList_Request,
+  GroupMessageEvent,
+} from "../../api/weshnet/protocoltypes";
+import { selectLastIdByKey, setLastId } from "../../store/slices/message";
+import { store } from "../../store/store";
+import { weshClient } from "../client";
+import { bytesFromString, stringFromBytes } from "../utils";
+
+export const subscribeMessages = async (groupPk: string) => {
```
WaDadidou, line 13: The same `suscribeMessages` is declared here : https://github.com/TERITORI/teritori-dapp/blob/45da028276b0157c5dd4eca50f1945e6b532a929/packages/weshnet/metadata/subscriber.ts#L16

sakul-budhathoki: Ok. I have removed the duplicate code.

PR #741, opened 2023-11-14
fix: Stick Article creation Toolbar at bottom

WaDadidou:
I wouldn't change the Toolbar emplacement in the container arborescence to avoid passing a ton of props from FeedNewArticleScreen to RichText.
But my solution seems messy


n0izn0iz:
We'll redo the article creation soon and @WaDadidou is not proud of this so closing

PR #747, opened 2023-11-15
feat: new article card

WaDadidou:
image

Mobile (360px) (Min width)
image

These buttons look better :
image
image

I sliced `SocialThreadCard` to make smaller and more reusable components
image

The user can now choose a Short Description to display on articles cards
image

"Cover image" is now named "Thumbnail image"
image


packages/components/socialFeed/NewsFeed/NewsFeed.type.ts
```diff
@@ -44,6 +48,22 @@ export interface SocialFeedMetadata {
   // openGraph?: OpenGraphType;
 }
 
+export const ZodSocialFeedArticleMetadata = z.object({
+  title: z.string(),
+  shortDescription: z.string(),
```
n0izn0iz, line 53: this should be optional since we don't have it in previous posts I think

packages/components/socialFeed/RichText/RichText.web.tsx
```diff
@@ -193,26 +188,18 @@ export const RichText: React.FC<RichTextProps> = ({
...
   useEffect(() => {
-    if (isTruncateNeeded) {
-      const contentState =
-        createStateFromHTML(initialValue).getCurrentContent();
-      const truncatedBlocks = contentState
-        .getBlocksAsArray()
-        .slice(0, NB_ROWS_SHOWN_IN_PREVIEW);
-      const truncatedState = ContentState.createFromBlockArray(truncatedBlocks);
-      const truncatedHtml = createHTMLFromState(truncatedState);
+    if (isArticleHTMLNeedsTruncate(initialValue, isPreview)) {
```
n0izn0iz, line 196: why not use isTruncateNeeded directly?

packages/components/socialFeed/NewsFeed/NewsFeed.type.ts
WaDadidou: If no `shortDescription`, the old way is used (Truncate the `message` html and show few lines of that)
See https://github.com/TERITORI/teritori-dapp/blob/9bda66ee2d4e8cfcb99b6a009e5e2be8e9ce3f67/packages/components/socialFeed/SocialCard/cards/SocialArticleCard.tsx#L95

packages/components/socialFeed/RichText/RichText.web.tsx
WaDadidou: It's  a mistake, fixed !

PR #756, opened 2023-11-20
feat(messenger): message ui components

sakul-budhathoki:
Run wesh daemon locally:
 - `go run ./go/cmd/weshd`
 - `go run ./go/cmd/weshd 4243`
 
 Route:
  - /message
 - /message?weshPort=4243


n0izn0iz:
there's some things I don't like in the code but we'll merge as-is since it's self-contained and we need to work on the desktop build

PR #757, opened 2023-11-21
feat(dAppStore): Move external apps into external sites

omniwired:
image


packages/screens/DAppStore/components/DAppStoreData.tsx
```diff
@@ -1,91 +0,0 @@
-import { useQuery } from "@tanstack/react-query";
```
omniwired, line 1: @n0izn0iz I'm sure you will love this one.

omniwired:
since the idea behind this change is to allow developers to add their apps in the dApp Store.
It makes no sense to keep the airtable implementation around.

packages/screens/DAppStore/query/getFromFile.ts
```diff
@@ -164,14 +162,14 @@ export function getAvailableApps(): dAppGroup {
...
         },
       },
     },
-    "coming-soon": {
-      id: "coming-soon",
-      groupName: "Coming soon dApps",
+    bookmarks: {
+      id: "bookmarks",
+      groupName: "External Apps",
       icon: otherAppsIcon,
```
WaDadidou, line 173: The icon for "coming-soon" and  "bookmarks" doesn't appear 
image

WaDadidou:
Some icons look weird in the Sidebar
image

packages/components/navigation/Sidebar.tsx
```diff
@@ -124,7 +124,13 @@ export const Sidebar: React.FC = () => {
           return (
             <SidebarButton
               key={item.id}
-              onPress={onRouteChange}
+              onPress={
+                route === "External"
```
WaDadidou, line 128: Maybe you can type `route` from `getFromFile.ts` with something like `"External" | "ComingSoon" | keyof RootStackParamList`

packages/components/ScreenContainer/index.tsx
```diff
@@ -67,6 +69,14 @@ export const ScreenContainer: React.FC<{
   forceNetworkKind,
   forceNetworkFeatures,
 }) => {
+  const dispatch = useAppDispatch();
```
n0izn0iz, line 72: can we move this to a component in App.tsx?
(see WalletSyncer component)

packages/components/navigation/Sidebar.tsx
omniwired: thanks for the code; 
Updated!

packages/components/ScreenContainer/index.tsx
omniwired: done; thanks for the feedback

PR #758, opened 2023-11-21
feat(app): build app

sakul-budhathoki:
This PR fixes all errors occurring before the homepage is displayed.


packages/components/OptimizedImage.tsx
```diff
@@ -65,13 +64,18 @@ const transformURI = (
...
+  }
+
   const isRelative = uri.startsWith("/");
   if (isRelative) {
     return uri;
   }
 
   // detect if raw CID
   try {
+    const { CID } = require("multiformats");
```
omniwired, line 78: how would this fix the issue?
I thought there is no react-native version of this library

omniwired: https://github.com/multiformats/js-multiformats/issues/276

Issue #776, opened 2023-11-27
Unable to connect wallet using leap mobile app dapp browser


dineshleap:
Unable to connect to https://app.teritori.com using leap mobile app dapp browser. Though the connection popup is approved the wallet state is not getting updated in the dapp.

Video ref:

https://github.com/TERITORI/teritori-dapp/assets/112848953/91af37dd-73c1-4493-9d20-398ee5d4d35a

PR #777, opened 2023-11-27
fix: connection issue leap mobile dapp browser (#776)

dineshleap:
### Issue
#776 

### Problem: 
The issue is happening because by the time the `useLeap` hook (which is responsible for checking leap object on load) tries to check for `window.leap` object using eventlistener on `load`, `window.leap` isn't available. So, the state of leap object is `undefined` throughout the app state which is causing the issue.

`packages/context/WalletsProvider/leap.ts`

### Solution:
Added a useEffect hook to check for selected wallet state and then trying to call the `load` eventlistener callback again which fixed the issue

Video ref:

https://github.com/TERITORI/teritori-dapp/assets/112848953/7c14afbb-3a28-44de-ab81-a5e1ea2eded0


omniwired (approved):

n0izn0iz:
thanks for the contribution :D

zxxma:
thanks @dineshleap 👌

PR #782, opened 2023-11-28
feat(tns): avatar and cover image upload previews

omniwired:


zxxma:
Bug report : 
I think there is something strange with pictures compression,
Looks like compressing to much and pixelizing (both avatar and header), not only the preview, but minted content.
Can you double-check ? 

(otherwise, it's nice! 👍)

zxxma:
Thanks for the fix 👌 
LGTM 👍

WaDadidou (approved):
LGTM :)

n0izn0iz:
When the image takes time to get replicated, the size breaks

WaDadidou:
> This can happen if a file is replicated for the first time: Before timeout, would be better to show a loader on a black background I think Screenshot 2023-11-30 at 20 02 14
> 
> After timeout Screenshot 2023-11-30 at 21 08 04
> 
> Also would be nice that the banner preview take the full width

Fixed

packages/components/teritoriNameService/NameDataForm.tsx
```diff
@@ -120,100 +124,106 @@ export const NameDataForm: React.FC<{
       />
       <TextInputCustom<Metadata>
```
clegirar, line 125: Already said that in an other PR for the same component, i don't ask to change it right now because it's not the purpose of the PR, but can you let a comment to say that it'll need to be refacto in a PR that will improve the `TextInput` uses ? (In every file you use it, to don't forget)

clegirar (commented):
Else i think everything is okay 👍

packages/components/teritoriNameService/NameDataForm.tsx
WaDadidou: Done.
Yes, TextInputCustom is too heavy

packages/components/OptimizedImage.tsx
```diff
@@ -22,12 +22,27 @@ export const OptimizedImage: React.FC<
...
           style={{
             width,
             height,
+            position: StyleSheet.flatten(other.style).position,
+            top: StyleSheet.flatten(other.style).top,
+            left: StyleSheet.flatten(other.style).left,
+            right: StyleSheet.flatten(other.style).right,
+            bottom: StyleSheet.flatten(other.style).bottom,
+            borderColor: StyleSheet.flatten(other.style).borderColor,
+            borderWidth: StyleSheet.flatten(other.style).borderWidth,
```
n0izn0iz, line 45: StyleSheet.flatten is an expensive operation, you should do it only once there
Especially in a component used that much

n0izn0iz (changes requested):

packages/components/OptimizedImage.tsx
WaDadidou: Fixed

PR #783, opened 2023-11-28
fix(feed): fix query in user feed

omniwired:
- Fixed query parameter name from "created_by" to "author_id" in the feed service.go file.
- This fixes an issue where the wrong query parameter was being used.
- The issue number is hotfix/profile-posts-not-working.


sakul-budhathoki (approved):
LGTM

PR #788, opened 2023-11-29
fix(button-capitalization): fix button text capitalization

omniwired:
This commit fixes the button text capitalization in the UploadMusicButton and PrimaryButton components. The textTransform CSS property is set to "capitalize" in both components.


packages/components/video/UploadVideoButton.tsx
```diff
@@ -58,6 +58,6 @@ const buttonContainerStyle: ViewStyle = {
 };
 const buttonTextStyle: TextStyle = {
   ...fontSemibold14,
-
+  textTransform: "capitalize",
```
WaDadidou, line 61: I'm not sure about this 
image

PR #792, opened 2023-11-29
feat: Make tabs of Social Feed as pages with proper url

WaDadidou:


clegirar (approved):
I think you can extract a component between all `FeedScreen` to be consistent, easier to change in the future, but components are not very long so it's not mandatory at all i guess (?)
So LGTM 👍

PR #804, opened 2023-12-11
feat(stake): randomize validators on redelegate

omniwired:


n0izn0iz:
this can show inactive validators first, not a good move IMO

PR #809, opened 2023-12-14
Control Wallet when user asking for transactions (not enough funds or not connected)

WaDadidou:
- New `WalletControlProvider`, that provides two modals
- Factorize Post creation (Article, Video, Music, simple Post) in `useFeedPosting`
- Use `WalletControlProvider` for :
    - Post creation (Article, Video, Music, simple Post)
    - React to a post (Like, Dislike, Emojis)
    - Mint a Name
    - Mint a NFT Collection
**==> More usages ?**
---

Wallet not connected
image

Not enough funds IBC currency
image

Not enough funds native currency 
image


clegirar (commented):
Maybe we can start to avoid long functions/components/files ? I think about `packages/hooks/feed/useMakePost.ts`, `packages/components/modals/ConnectWalletModal.tsx` and `packages/components/modals/NotEnoughFundsModal.tsx`
We can even define a limit maybe ? Without be dumb, the goal stay to be readable

packages/hooks/feed/useFeedPosting.ts
```diff
@@ -133,30 +130,24 @@ export const useFeedPosting = (
...
+  const { makeSimplePost, makeVideo, makeTrack, makeArticle, isLoading } =
+    useMakePost({ networkId, userId, category, processFeedPosting });
+
   return {
-    canPayForPost,
-    makePost,
+    makeSimplePost,
+    makeArticle,
+    makeTrack,
+    makeVideo,
```
n0izn0iz, line 143: like we discussed, it's not a good idea to take a category in input and output multiple callbacks, this allows to misuse the function

n0izn0iz (changes requested):

packages/hooks/feed/useFeedPrettyBalanceFee.ts
```diff
@@ -0,0 +1,36 @@
...
+import { useBalances } from "../useBalances";
+import { useSelectedNetworkId } from "../useSelectedNetwork";
+
+interface UseFeedPrettyBalanceFeeParams {
+  userId?: string;
+  category: PostCategory;
+}
+
+export const useFeedPrettyBalanceFee = ({
+  userId,
```
n0izn0iz, line 14: this will most likely behave wrongly when there is no wallet connected, either split the balance and publishing fee in two hooks or take an additional networkId param

n0izn0iz (changes requested):

packages/hooks/feed/useFeedPrettyBalanceFee.ts
WaDadidou: Done

packages/hooks/feed/useFeedPosting.ts
WaDadidou: ======> I will revert the part "refacto useFeedPosting"

PR #824, opened 2023-12-18
fix: mobile title feed screens

WaDadidou:
### Before

image

### After

image

### + Simplify ScreenContainer props for Feed tab screens


packages/screens/Feed/FeedScreen.tsx
```diff
@@ -8,29 +8,62 @@ import { VideosFeedScreen } from "./VideosFeedScreen";
...
+    footerChildren: <></>,
+    headerChildren: <BrandText>Social Feed</BrandText>,
+  };
 
   switch (params?.tab) {
     case "music":
-      return <MusicFeedScreen />;
+      return (
+        <MusicFeedScreen
+          screenContainerOtherProps={screenContainerOtherProps}
```
clegirar, line 42: If the props are all the same maybe you can create a base of `FeedScreen` that is juste a `ScreenContainer` with this props that just take a children ?
It's a proposal of course, and we can do it in an other PR 👍

WaDadidou: Ah yes of course

clegirar (approved):
LGTM 👍

PR #825, opened 2023-12-18
fix: remove useless scroll on GridList (Feed Music, NFTs, etc..)

WaDadidou:
### Before
image


clegirar (approved):
LGTM 👍

n0izn0iz:
if you don't put a fixed height, it will load the whole list, 3K nft for the riot example this will auto ddos us

WaDadidou:
CLOSED : No issue, it's ok

n0izn0iz:
maybe remove the footers in UPP for better UX though and maybe at a later time, use the FlatList header prop to prevent nested scrolling

PR #876, opened 2024-01-17
Feat/map confirm modal

thomarnauld:
Hi,
Show Map Modal before send post is done.
The senario is : 
1- if adress input is empty the 'add location' button is disables.
2- if user entrer an address a function return a list of locations corresponding to the address
3- il user select a location the map update and the input placeholder is changed to location label.
 
Screenshot 2024-01-17 at 11 11 08

I fixed the bug on children who are not well centered in the boxes


thomarnauld:
Create a new PR here >>> https://github.com/TERITORI/teritori-dapp/pull/898. This PR is deprecated

PR #883, opened 2024-01-23
fix: replaces useDropdowns by useClickOutside

sakul-budhathoki:
**Problem:**
Due to the use of Pressable on `DropdownsContextProvider`, it blocks scroll of `ScrollView` on mobile app (Android & iOS)

**Solution:**
Replace `useDropdowns` by [react-native-click-outside](https://github.com/jakex7/react-native-click-outside)


App.tsx
```diff
@@ -114,7 +114,7 @@ export default function App() {
                             </SearchBarContextProvider>
                           </WalletControlContextProvider>
                         </WalletsProvider>
-                      </DropdownsContextProvider>
+                      </ClickOutsideProvider>
```
omniwired, line 117: How about "DropdownsClickOutsideProvider"; maybe is too long, but I feel that with the new name the original intent "Dropdown" got lost.

sakul-budhathoki: Thanks for the idea, yes I agree we can achieve new changes without renames. 
I can use:
~ClickOutsideProvider~ -> DropdownsContextProvider
~useClickOutside~ -> useDropdowns

So, we will have naming conventions as before, and it will be easier for devs to use this.

clegirar: Yep totally agree with that too 👍

packages/components/nfts/NFTView.tsx
```diff
@@ -118,10 +114,10 @@ export const NFTView: React.FC<{
 
 const NFTViewContent: React.FC<{
   nft: NFT;
-  dropdownRef: RefObject<View>;
+
```
clegirar, line 117: This line is empty right ? If you can delete it 👍

clegirar (changes requested):
Else found that good 👍 Thanks for that 👍

PR #834, opened 2023-12-19
fix: updated legacy boxes

thomarnauld:
Hello, I created this new branch to validate the CI. I commented out the variables not used in the code (Legacy) in case we need them in the future. I will delete the comments once this first test is validated. THANKS

PR #897, opened 2024-01-26
chore: first replace of legacies boxes

thomarnauld:
Created smaller and cleaned PR based on https://github.com/TERITORI/teritori-dapp/pull/834


packages/components/boxes/Box.tsx
```diff
@@ -53,6 +53,7 @@ export type BoxStyle = Pick<
   | "top"
   | "bottom"
   | "flex"
+  | "zIndex"
```
n0izn0iz, line 56: zIndex should be passed to the outer component IMO

n0izn0iz (changes requested):

packages/components/boxes/Box.tsx
clegirar: Yep i agree

PR #919, opened 2024-02-02
chore: cache wesh framework on CI

n0izn0iz:


.github/workflows/mobile-build-ios.yml
```diff
@@ -34,6 +34,12 @@ jobs:
       - name: Fix gitignore
         run: npx tsx ./packages/scripts/app-build/fixGitignore.ts
 
+      - name: Restore wesh framework from cache if possible
```
sakul-budhathoki, line 37: @n0izn0iz 
For the first time, it won't find the cache.
Then, it will generate weshframework.
Will it cache the newly generated weshframework?

We should add the cache again for the first build after the weshframework build.
Not sure how actions/cache@v4 works; sorry if I mistaken.

n0izn0iz: this step adds a post action that save the files at `path`

PR #935, opened 2024-02-07
feat: eas submit on push to main

sakul-budhathoki:


packages/scripts/app-build/bumpBuildNumber.ts
```diff
@@ -0,0 +1,22 @@
+import fs from "fs";
+
+const FILE_PATH = "./app.config.js";
+const buildNumber = process.argv[2];
+
+fs.readFile(FILE_PATH, "utf8", (err, data) => {
+  if (err) {
+    console.error("Error reading app.config.js", err);
+    return;
```
n0izn0iz, line 9:
```suggestion
    process.exit(1);
```
so the ci also fails if this fails

packages/scripts/app-build/bumpBuildNumber.ts
```diff
@@ -0,0 +1,22 @@
...
+  }
+  const regex = /buildNumber: "(\d+)"/;
+
+  const updatedData = data.replace(regex, 'buildNumber: "' + buildNumber + '"');
+
+  fs.writeFile(FILE_PATH, updatedData, "utf8", (err) => {
+    if (err) {
+      console.error("Error updating app.config.js:", err);
```
n0izn0iz, line 17:
```suggestion
      console.error("Error updating app.config.js:", err);
      process.exit(1);
```

n0izn0iz: same

n0izn0iz (changes requested):

PR #946, opened 2024-02-12
chore: Use shorter imports

omniwired:


sakul-budhathoki (approved):
LGTM. Thanks.
I was thinking of doing it using regexp script to auto update; but your IDE makes it easier to handle such things.

Later; we might need to do similar for `assets` too. 
Norman has already configured for assets path on `tsconfig.json`
` "@/assets/*": ["./assets/*"],`

PR #953, opened 2024-02-13
fix(native-wallet): gas estimation adjustment and button title

omniwired:
- In SendingToriScreen, fix the gas estimation by adding 100,000 to the estimated gas value.
- Update the button title in SendingToriScreen to "Sending" when the transaction is in process.
- Set the isInProcess state to true when the transaction is being signed and broadcasted.


sakul-budhathoki (dismissed):
LGTM. Thanks.

clegirar (dismissed):
Code LGTM 👍

PR #954, opened 2024-02-14
feat: Wallet, send UI improvements

omniwired:


packages/screens/Mini/components/MiniAddressSelectorDropdown.tsx
clegirar, line 1: On this file, if you think that `useSelector` will be used later on other screens, you can create specific hooks ?

sakul-budhathoki (dismissed):
LGTM

packages/screens/Mini/components/MiniAddressSelectorDropdown.tsx
omniwired: I don't think it makes sense now.
It's an easy-to-follow one-liner, which I believe is not necessary.

clegirar (approved):
Code seems good 👍 Don't really like the second commit name but i'll not nitpick on that. LGTM 👍

omniwired:
> Code seems good 👍 Don't really like the second commit name but i'll not nitpick on that. LGTM 👍

it will be squashed

sakul-budhathoki (approved):

PR #957, opened 2024-02-14
feat: allow edit profile directly on the user profile

VishalRakholiya-iView:


omniwired (approved):

packages/utils/navigation.ts
```diff
@@ -21,6 +21,7 @@ export type RootStackParamList = {
   UserPublicProfile: {
     id: string;
     tab?: keyof typeof uppTabItems;
+    isEditProfile?: boolean;
```
clegirar, line 24: Why you add this but you didn't use it in the screen ?
Please remove it if you don't use it

packages/screens/UserPublicProfile/tabScreens/PostsUPPScreen.tsx
```diff
@@ -11,6 +11,7 @@ import { useIsDAO } from "@/hooks/cosmwasm/useCosmWasmContractInfo";
 import { useIsDAOMember } from "@/hooks/dao/useDAOMember";
 import { useNSUserInfo } from "@/hooks/useNSUserInfo";
 import { parseUserId } from "@/networks";
+// import { TNSUpdateNameScreen } from "@/screens/TeritoriNameService/TNSUpdateNameScreen";
```
clegirar, line 14: Useless comment here, you can delete it too

clegirar (changes requested):

clegirar (approved):
LGTM 👍

PR #959, opened 2024-02-14
feat: add native signer to the mobile app

omniwired:
This add the native wallet signer into the app.

image


packages/utils/feed/queries.ts
```diff
@@ -14,14 +11,17 @@ import {
...
 export const getAvailableFreePost = async (userId: string) => {
   const [network, userAddress] = parseUserId(userId);
 
   if (!userAddress || network?.kind !== NetworkKind.Cosmos) {
     return;
   }
 
-  const client = await signingSocialFeedClient({
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  const client = await useSigningSocialFeedClient({
```
n0izn0iz, line 24: this eslint error is pointing at a real bug, if you use a hook conditionally, it will crash the render (the condition is the if with return up

n0izn0iz, line 24: move this in hooks, it should resolve the imports restriction problem also

packages/hooks/feed/useSocialReactions.ts
```diff
@@ -50,7 +50,7 @@ export const useSocialReactions = ({
       },
     });
   const cosmosReaction = async (emoji: string, walletAddress: string) => {
-    const client = await signingSocialFeedClient({
+    const client = await useSigningSocialFeedClient({
```
n0izn0iz, line 53: the linter error is pretty explicit, need to be fixed

packages/hooks/feed/useFeedPosting.ts
```diff
@@ -130,7 +130,8 @@ export const useFeedPosting = (
         } else {
           setStep(feedPostingStep(FeedPostingStepId.POSTING));
 
-          const client = await signingSocialFeedClient({
+          // eslint-disable-next-line react-hooks/rules-of-hooks
```
n0izn0iz, line 133: don't disable rules of hook, see other comment about this

packages/components/socialFeed/SocialActions/TipModal.tsx
```diff
@@ -106,7 +107,8 @@ export const TipModal: React.FC<{
         setToastError({ title: "Tip failed", message: err.message });
       }
     } else {
-      const client = await signingSocialFeedClient({
+      // eslint-disable-next-line react-hooks/rules-of-hooks
```
n0izn0iz, line 110: same thing

n0izn0iz (changes requested):

packages/hooks/feed/useSocialFeedClient.ts
```diff
@@ -0,0 +1,48 @@
...
+  const selectedWallet = useSelectedNativeWallet();
+
+  if (cachedSigningClients[cacheKey]) {
+    return cachedSigningClients[cacheKey];
+  } else {
+    let signingComswasmClient;
+    if (Platform.OS === "web") {
+      signingComswasmClient = await getKeplrSigningCosmWasmClient(networkId);
+    } else {
+      signingComswasmClient = await getNativeSigner(selectedWallet!);
```
n0izn0iz, line 36: you shouldn't allow exception in hooks

n0izn0iz (changes requested):

packages/utils/wallet/getNativeSigner.ts
```diff
@@ -1,16 +1,17 @@
-import { SigningStargateClient } from "@cosmjs/stargate";
+import { SigningCosmWasmClient } from "@cosmjs/cosmwasm-stargate";
 
 import { getNativeWallet } from "./getNativeWallet";
 
 import { mustGetCosmosNetwork } from "@/networks";
+// eslint-disable-next-line no-restricted-imports
```
omniwired, line 6: @n0izn0iz I want to use this type, should I put this in another file, so I can use them in both places? (here and slices)

n0izn0iz: put StoreWallet in utils/types IMO

n0izn0iz (changes requested):

n0izn0iz (changes requested):
hooks should be in hooks dir

PR #960, opened 2024-02-14
fix: mini navigation

sachin-into:
1. moved normal navigation and mini navigation under same stack navigation
2. fixed dApps onPress navigation
3. added chat\feed\wallets button at bottom of dApps screen


omniwired (approved):

packages/components/navigation/MiniModeNavigator.tsx
```diff
@@ -94,9 +87,9 @@ const MainTab = () => {
   );
 };
 
-export const MiniModeNavigator: React.FC = () => {
+export const MiniModeNavigator = () => {
```
n0izn0iz, line 90:
```suggestion
export const getMiniModeScreens = () => {
```

n0izn0iz: the other name is confusing as the capitalized names are used for components and this is not a component but rather a function that returns jsx, also it's not a navigator anymore

n0izn0iz (changes requested):

omniwired (approved):

PR #969, opened 2024-02-16
feat: remove a single wallet

omniwired:


clegirar (dismissed):
Code seems good.

packages/screens/Mini/Wallet/TokenScreen.tsx
```diff
@@ -33,15 +34,16 @@ import {
...
 
   const selectedWallet = useSelectedNativeWallet();
   if (!selectedWallet) {
-    dispatch(setSelectedNativeWalletIndex(wallets[0].index));
+    if (wallets.length !== 0) {
+      dispatch(setSelectedNativeWalletIndex(wallets[0].index));
+    } else {
+      navigation.navigate("NativeWallet");
+    }
   }
```
n0izn0iz, line 47: calling dispatch and navigate in render function is very bad and can cause all sorts of crashes and weird behaviors, it should be in an effect or action callback

n0izn0iz: a component function should be a "pure" function that transforms it's props into jsx, side-effects should only happen in effects (useEffect) and callbacks (onPress, onTextChange, etc..)

n0izn0iz (changes requested):

packages/screens/Mini/Wallet/TokenScreen.tsx
n0izn0iz: see https://react.dev/learn/render-and-commit

omniwired: is this related https://react.dev/learn/you-might-not-need-an-effect ?

n0izn0iz: well, this is the counterpart, people often overuse effects and memos when it's not needed, both are interesting but it's not the related to the problems on these lines

packages/screens/Mini/Profile/AccountDetailsScreen.tsx
```diff
@@ -30,16 +36,43 @@ export const AccountDetailsScreen: ScreenFC<"MiniAccountDetails"> = ({
...
+      dispatch(updateWallet({ ...wallet, name: text }));
     }
     setAccountName(text || "");
   };
 
+  const onResetPress = () => {
+    if (wallet) {
+      Alert.alert(
+        "Are you sure?",
+        "This action will remove the wallet from your device. You can always import it back.",
```
n0izn0iz, line 55:
```suggestion
        "This action will remove the wallet from your device. If you didn't save the seed, this wallet will be lost forever! If you saved the seed, you will be able to import it back later.",
```

PR #971, opened 2024-02-16
feat(mini-mode): comments input

sachin-into:
1. modified useIpfs hook to upload files in mini mode 
2. created MiniCommentInput (similar to NewsfeedInput but for comment only); has different file browsing packages supporting mobile for image/video/audio selection


omniwired (approved):

packages/screens/Mini/components/MiniCommentInput.tsx
clegirar, line 1: I really think that have a file of 680 lines of code is so hard to read/maintain/scale/change, maybe you can split it in different features like camera, emoji, files etc. ?

sachin-into: ok, sure

clegirar: If you have other ideas on how to split that don't hesitate to make a proposal

omniwired (approved):

packages/hooks/useIpfs.ts
```diff
@@ -22,6 +24,7 @@ interface IPFSUploadProgress {
 export interface PinataFileProps {
   file: LocalFileData;
   pinataJWTKey: string;
+  mode?: AppMode;
```
n0izn0iz, line 27: I don't understand why this hook need to be aware of the app mode

packages/hooks/useIpfs.ts
```diff
@@ -33,10 +36,20 @@ export const useIpfs = () => {
     async ({
       file,
       pinataJWTKey,
+      mode,
     }: PinataFileProps): Promise<string | undefined> => {
       try {
         const formData = new FormData();
-        formData.append("file", file.file);
+        if (mode === "mini") {
+          //@ts-expect-error: description instead of adding file when adding url in formdata file upload to pinata
```
n0izn0iz, line 44: I don't understand the comment

packages/hooks/useIpfs.ts
sachin-55: While uploading file from app instead of converting selected file to  File type similar to WEB, I just used file URL to create formData and upload it to pinata. So, just to check if we are uploading from web or app I added that appmode

packages/hooks/useIpfs.ts
sachin-55: As I state in above comment, to create formData for app side, I did it in that way.

packages/hooks/useIpfs.ts
n0izn0iz: okay, app mode is not related to mobile or desktop, it's the mode of the ui, you could use mini-mode on desktop, btw mini means minimalist, not small screen

n0izn0iz: use Platform.OS in here directly if possible and remove the new prop

sachin-55: ok, I didn't knew that. Will do

sachin-55: I replaced app mode check with Platform.OS check

PR #978, opened 2024-02-19
feat: change input of name TNS register

VishalRakholiya-iView:


packages/components/modals/teritoriNameService/TNSNameFinderModal.tsx
clegirar, line 1: You don't create a specific component, you just copy what they done in `CreateDAOSection.tsx` but it's better to have an input component that check available names, that call the `TextInputCustom` etc. and called this component in `CreateDAOSection` and `TNSNameFinderModal`

clegirar (changes requested):

packages/components/modals/teritoriNameService/TNSNameFinderModal.tsx
VishalRakholiya-iView: you mean use `TextInputCustom` in both file (`CreateDAOSection` and `TNSNameFinderModal`)?

and write logic in `TextInputCustom` ?

n0izn0iz (changes requested):
I agree with clemsss, please extract the component

packages/components/modals/teritoriNameService/TNSNameFinderModal.tsx
```diff
@@ -105,6 +110,19 @@ export const TNSNameFinderModal: React.FC<{
...
+  const networkId = network?.id;
+  const cosmosNetwork = getCosmosNetwork(networkId);
+  const normalizedTokenId = (
+    name + cosmosNetwork?.nameServiceTLD || ""
+  ).toLowerCase();
+
+  const tokenId = name + cosmosNetwork?.nameServiceTLD || "";
+
+  const { nsMintPrice: price } = useNSMintPrice(networkId, normalizedTokenId);
+  const { nameAvailable, loading } = useNSMintAvailability(networkId, tokenId);
```
n0izn0iz, line 124: all this should be inside the `AvailableNamesInput` component

n0izn0iz (changes requested):

packages/components/modals/teritoriNameService/TNSNameFinderModal.tsx
clegirar: Yep you have to move the logic too in the component, if it's possible, but i think it is

clegirar (changes requested):
When you moved the logic in the new component, please clean comments in the code

packages/screens/Organizations/components/CreateDAOSection.tsx
```diff
@@ -7,19 +7,20 @@ import { RadioDescriptionSelector } from "./RadioDescriptionSelector";
...
 import { useNSAvailability } from "@/hooks/useNSAvailability";
 import { useSelectedNetworkInfo } from "@/hooks/useSelectedNetwork";
-import { getCosmosNetwork, NetworkKind } from "@/networks";
+import { NetworkKind } from "@/networks";
 import {
   neutral33,
   neutral77,
-  primaryColor,
-  redDefault,
+  // primaryColor,
```
clegirar, line 19: Just remove this comments

clegirar (changes requested):

clegirar:
And please pass the CI, don't export `TextInputCustomProps`

clegirar:
The only thing i noticed in the UI is the equivalent in `$`
Screenshot 2024-02-22 132129
This one is from `create-org` screen
Screenshot 2024-02-22 132139
This one is from `tns`

Please add the equivalent in dollars in the custom component you created 👍

clegirar (approved):
LGTM 👍

PR #984, opened 2024-02-21
feat(wallet): wallet add derivation paths

omniwired:


packages/hooks/wallet/getNativeWallet.ts
```diff
@@ -1,10 +1,15 @@
 import { Secp256k1HdWallet } from "@cosmjs/amino";
+import { stringToPath } from "@cosmjs/amino/node_modules/@cosmjs/crypto/build/slip10";
```
omniwired, line 2: I had to do this. To make tsc stop complaining
Do you know of any other options?

PR #993, opened 2024-02-23
feat: weshnet chat flow

sujal-into:
- Chat force activation
- Chat get started flow


packages/components/buttons/ToggleButton.tsx
```diff
@@ -17,30 +15,34 @@ interface ToggleButtonProps extends SwitchProps {
...
+      return neutral99;
+    }
+
+    return isActive ? blueDefault : neutral33;
+  };
+
   return (
     <Switch
-      // @ts-expect-error: the active thumb color is a weird green that can't be changed without "activeThumbColor"
-      activeThumbColor={secondaryColor}
```
n0izn0iz, line 37: this change breaks the thumb color, like written in the comment
Screenshot 2024-02-25 at 10 09 51

n0izn0iz (changes requested):

packages/components/buttons/ToggleButton.tsx
sakul-budhathoki: Thanks, I didn't realize the issue was on web.

PR #997, opened 2024-02-23
fix(wesh): remove wesh autolinking

omniwired:


package.json
```diff
@@ -200,10 +200,5 @@
     "@axelar-network/axelarjs-types@^0.33.0": "patch:@axelar-network/axelarjs-types@npm%3A0.33.0#./.yarn/patches/@axelar-network-axelarjs-types-npm-0.33.0-5d36465d11.patch"
   },
   "private": true,
-  "packageManager": "yarn@3.5.0",
-  "expo": {
```
sakul-budhathoki, line 204: There are two ways to link the native modules on Expo:
-  add native modules in ./modules folder then expo is autolink it by default.
- another adding autolinking config on package.json

Previously I used ./modules folder for weshd; it had some issues on the linking on github ci so I have used package.json method.

Doing this; it will lead to error due to below code
`const WeshnetModule = require("../../weshd")` in packages/weshnet/services.ts

omniwired:
Thanks @sakul-budhathoki for the clarification.
This PR is just in case we need to keep moving with the app development and testing without wesh. (As least temporarily)

n0izn0iz:
obsolete

PR #999, opened 2024-02-26
fix: dismiss keyboard on MutiStepScreenContainer and BlurScreenContainer

sachin-into:
1. hiding keyboard on clicking on screen other than input by wrapping component by ScrollView with scroll disabled, 
2. added focus on input while clicking on seed input container on Import seed screen


omniwired (dismissed):
🚀

PR #1003, opened 2024-02-27
fix: scroll on FeedMusicList and FeedVideosList

sachin-into:


packages/components/video/FeedVideosList.tsx
```diff
@@ -111,6 +117,7 @@ export const FeedVideosList: React.FC<{
             />
           )}
           onEndReached={onEndReached}
+          noFixedHeight
```
n0izn0iz, line 120: this completely breaks pagination, ie loading all possible video/music, be careful, for now it's ok but as more content is added, this will be problematic

sachin-into: So, leave it for now and fix later, when there will be problem.

n0izn0iz: there is probably another way and I'd prefer that we don't add technical debt
please, at the very least, add a comment about this in the code

sachin-into: ok

omniwired (approved):
🚀Now 🛠️ later

PR #1009, opened 2024-02-28
fix: index nft transfered by another contract

hthieu1110:


go/internal/substreams/ethereum/handlers/nft.go
```diff
@@ -20,6 +21,28 @@ type TransferInput struct {
...
+	if err := h.indexerDB.Where("id", nftId).First(&nft).Error; err != nil {
+		return errors.Wrap(err, "failed to get nft")
+	}
+
+	nft.OwnerID = h.network.UserID(to)
+	if err := h.indexerDB.Save(&nft).Error; err != nil {
+		return errors.Wrap(err, "failed to update nft owner")
+	}
+
+	// NOTE: This is used only to handle transfers made by another contract so we dont track activities
```
n0izn0iz, line 41: why? in the teritori indexer, we still add to activity in this case (transfer activity)
can be transfer to contract or user, make sense to add it to activity

n0izn0iz (approved):
LGTM

PR #1015, opened 2024-02-28
chore: build framwork based on arch

sakul-budhathoki:


Makefile
```diff
@@ -34,6 +34,15 @@ P2E_DOCKER_IMAGE=$(DOCKER_REGISTRY)/p2e-update-leaderboard:$(shell git rev-parse
 FEED_DOCKER_IMAGE=$(DOCKER_REGISTRY)/feed-clean-pinata-keys:$(shell git rev-parse --short HEAD)
 MULTISIG_DOCKER_IMAGE=$(DOCKER_REGISTRY)/cosmos-multisig-backend:$(shell git rev-parse --short HEAD)
 
+
+ARCH := $(shell uname -m)
+
+ifeq ($(ARCH),x86_64)
```
omniwired, line 40: question:
does it _ever_ make sense to build it for x86, given that both Android and iOS are ARM?

sakul-budhathoki: It is not for the target (Android and iOS) but for the source (where the framework will be generated) Macbook with Intel chip or M chips.
But I may be wrong on this; I created this so that the dev doesn't have to set the GOARCH manually.

omniwired: understood, in that case it should be ARM only.
As this code never runs native outside of a phone.
In the case these compilation is needed for desktop.... hmm, in that case IDK.

omniwired (approved):

PR #1016, opened 2024-02-29
fix: scroll on mobile

sakul-budhathoki:


omniwired (approved):
tested on simulator and real device (iPhone 14 pro)
It works well.
Thanks!

clegirar (approved):
Tested on Android.
It works well 👍 
LGTM 👍

PR #1017, opened 2024-02-29
fix: image and video being shown small in mini mode

sachin-into:


packages/components/FilePreview/FilesPreviewsContainer.tsx
```diff
@@ -27,6 +28,7 @@ export const FilesPreviewsContainer: React.FC<FilePreviewContainerProps> = ({
   onDelete,
   onDeleteGIF,
   onAudioUpdate,
+  showSmallPreview = false,
```
omniwired, line 31: I don't think this is ever set to anything other than false.
Please help me understand why this prop is needed (and where is being used)

packages/screens/Mini/components/MiniCommentInput/CommentTextInput.tsx
```diff
@@ -185,6 +185,7 @@ export const CommentTextInput = React.forwardRef<MiniCommentInputHandle, Props>(
                     );
                   }
                 }}
+                showSmallPreview
```
sachin-into, line 188: In mini comment I used showSmallPreview flag to decrease height of selected image and video as using height as 400 with on screen keyboard would mess with UI

packages/components/FilePreview/FilesPreviewsContainer.tsx
sachin-into: yeah it is mostly false, but in mini comment input it is true

omniwired (approved):

PR #1024, opened 2024-03-04
wi: init cosmwasm contract

hthieu1110:


Makefile
```diff
@@ -3,7 +3,7 @@ BUNKER_MINTER_PACKAGE=teritori-bunker-minter
 GO?=go
 GOFMT?=$(shell $(GO) env GOROOT)/bin/gofmt
 
-COSMWASM_CONTRACTS_DIR=cosmwasm-contracts
+COSMWASM_CONTRACTS_DIR=rust
```
n0izn0iz, line 6:
```suggestion
COSMWASM_CONTRACTS_DIR=rust/cw-contracts
```

so you don't need the filter bellow
`if [[ $${indir} == "rust/apipb" ]]; then continue; fi ; \`

api/launchpad/v1/launchpad.proto
```diff
@@ -0,0 +1,128 @@
...
+  uint64 start_time = 34;
+
+  // Whitelist minting --------------------------
+  repeated string whitelist_addresses = 35;
+
+  uint64 whitelist_unit_price = 36;
+  string whitelist_limit_per_address = 37;
+  uint32 whitelist_member_limit = 38;
+  uint64 whitelist_start_time = 39;
+  uint64 whitelist_end_time = 40;
```
n0izn0iz, line 118: this should be an array of whitelist periods, we already have this in some collections

packages/networks/types.ts
```diff
@@ -25,6 +25,7 @@ interface NetworkInfoBase {
   contractExplorer: string;
   testnet: boolean;
   backendEndpoint: string;
+  launchpadEndpoint: string;
```
n0izn0iz, line 28: can we use a feature object instead?

- the base network object is already too heavy
- this allows to check for presence of all required params at once

packages/networks/types.ts
```diff
@@ -72,6 +73,7 @@ export type CosmosNetworkInfo = NetworkInfoBase & {
   daoVotingCw4CodeId?: number;
   daoFactoryContractAddress?: string;
   coreDAOAddress?: string;
+  launchpadContractAddress?: string;
```
n0izn0iz, line 76: can we use a feature object?

n0izn0iz (changes requested):

api/launchpad/v1/launchpad.proto
```diff
@@ -36,12 +36,12 @@ message CollectionsByCreatorResponse {
...
-  int32 limit = 3;
-  int32 offset = 4;
-  Sort sort = 5;
-  SortDirection sort_direction = 6;
+  string network_id = 1;
+  int32 limit = 2;
+  int32 offset = 3;
+  Sort sort = 4;
+  SortDirection sort_direction = 5;
+//  TODO: user authentication (Member of the admin DAO)
```
n0izn0iz, line 44: no need to add todo there, auth will be handled by a grpc middleware and grpc headers

packages/hooks/launchpad/useCompleteCollection.ts
```diff
@@ -113,7 +114,7 @@ export const useCompleteCollection = () => {
             );
 
             metadatas.push({
-              image: image.hash,
+              image: CID.parse(image.url).toV0().toString(),
```
n0izn0iz, line 117: you didn't read what I said, this should be an ipfs uri, not a raw cid

and I don't understand why you need to use a v0 cid

please see https://docs.opensea.io/docs/metadata-standards#ipfs-and-arweave-uris

n0izn0iz (changes requested):

go/pkg/launchpad/utils.go
```diff
@@ -146,5 +146,8 @@ func (m *Metadata) ToJSONB() interface{} {
 }
 
 func extractCID(imageURL string) string {
+	if len(imageURL) >= 7 && imageURL[:7] == "ipfs://" {
```
n0izn0iz, line 149: idiomatic golang:

```suggestion
	if strings.HasPrefix(imageURL, "ipfs://") {
```

packages/components/hub/HubLanding.tsx
```diff
@@ -37,8 +38,19 @@ export const HubLanding: React.FC = () => {
       <View style={{ flex: 1 }}>
         {!!banner && (
           <>
-            <SpacerColumn size={7} />
-            <HubBanner banner={banner} />
+            <Link to={banner?.url} style={{ width: "100%", maxHeight: 500 }}>
+              <OptimizedImage
+                sourceURI={banner?.image}
+                width={width}
```
n0izn0iz, line 44: this value should not be dynamic, the resizing proxy will create and cache an image for every new width passed

n0izn0iz (changes requested):

packages/components/hub/HubLanding.tsx
WaDadidou: There is no change on `HubLanding.tsx` on this PR (I rolled back), it's out of scope

PR #1026, opened 2024-03-05
feat: mini create post

sachin-into:
Replaced create post input from normal mode to mini input for creating post


packages/screens/Mini/components/MiniCommentInput/CommentTextInput.tsx
```diff
@@ -95,7 +98,7 @@ export const CommentTextInput = React.forwardRef<MiniCommentInputHandle, Props>(
                 }
                 onBlur={onBlur}
                 onFocus={onFocus}
-                placeholder="Hey yo! Write your comment"
+                placeholder={`Hey yo! ${type === "comment" ? "Write your comment" : "Post something"}`}
```
omniwired, line 101: I know is not related to this PR.
I think we should change the "Hey yo" for something else.

omniwired (approved):

PR #1064, opened 2024-03-07
fix: image overflow when viewed

### LINKED ISSUE #1054: Mobile: feed Preview photo layout issue
- Is too low.
- The image overflows

IMG_2531

sujal-into:
fixes #1054

### DIFF
--- a/packages/components/FilePreview/ImagesFullViewModal.tsx
+++ b/packages/components/FilePreview/ImagesFullViewModal.tsx
@@ -1,5 +1,10 @@
 import React, { useState } from "react";
-import { useWindowDimensions, View, TouchableOpacity } from "react-native";
+import {
+  useWindowDimensions,
+  View,
+  TouchableOpacity,
+  Platform,
+} from "react-native";
 import { SvgProps } from "react-native-svg";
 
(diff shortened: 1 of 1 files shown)


packages/components/FilePreview/ImagesFullViewModal.tsx
```diff
@@ -52,6 +57,9 @@ export const ImagesFullViewModal: React.FC<ImageFullViewModalProps> = ({
 }) => {
   const [localActiveIndex, setLocalActiveIndex] = useState(activeIndex);
   const { height, width } = useWindowDimensions();
+
+  const imageHeight = Platform.OS !== "web" ? height - 200 : height - 100;
```
omniwired, line 61: this one feels quite "magical" and opaque.
Why are exactly these values?

sakul-budhathoki: @omniwired  
On web 100px is used to set gap of 50px on top and 50px on bottom of the image. 
On app;  200 is required due to  insets. 
I will check if this can be removed completely.

omniwired (approved):
🎉

Issue #1036, opened 2024-03-07
Mobile: Video playing not smooth


omniwired:
https://github.com/TERITORI/teritori-dapp/assets/322761/c25de701-c976-44d0-92fd-e3e7724d8c01

sujal-into:
I will be working on this.

PR #1068, opened 2024-03-07
fix: video player fixed

sujal-into:
fixes: #1036


packages/components/mediaPlayer/MediaPlayerVideo.tsx
```diff
@@ -79,8 +79,10 @@ export const MediaPlayerVideo: FC<MediaPlayerVideoProps> = ({
   // Plug or not the playbackStatus from MediaPLayerProvider
   const statusToUse = useMemo(
     () => (isInMediaPlayer ? playbackStatus : localStatus),
-    [isInMediaPlayer, playbackStatus, localStatus],
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
omniwired, line 82: please let me understand why this is necessary?

sujal-into: Do you mean the comment? The comments are just to disable the warning which is sent by react if we not set the variable that is used inside the useMemo.   The variables playbackStatus and localStatus  in the dependency array is continously changing which updates the statusToUse continously, as a result the video component is rendering again and again, which plays and pause the video automatically, which is the reason for our video not playing.

omniwired: @n0izn0iz seems like a valid reason."

omniwired (approved):

packages/components/mediaPlayer/MediaPlayerVideo.tsx
n0izn0iz: I recommend not to ignore the error and refactor to not trigger

n0izn0iz: actually this memo and the one above should not be needed and are probably slowing down the render

n0izn0iz (changes requested):

n0izn0iz:
maybe you need to store  playbackStatus &/or localStatus in a useRef instead?

SujalLama:
> maybe you need to store playbackStatus &/or localStatus in a useRef instead?

thanks for the suggestion, i will look into it

sujal-into:
What I did?
- Break down the MediaPlayerVideo into small components
    - Moved the Video part of MediaPlayerVideo into ExpoAvVideo
    - Moved the ExpoAvVideo’s PlayerController into MediaPlayerController
- Got rid of unnecessary code that wasn’t helping
    - Removed statusToUse 
        - We didn’t need the playbackStatus only local Status that’s why I removed the playbackStatus and relocated the localStatus to Media pLayer controller
        - StatusToUse : Populate two values (localStatus or playbackStatus) based on isInMediaPlayer
            - When isInMediaPlayer = true, StatusToUse is always undefined (which is the value returned from playbackStatus) 
            - When isInMediaPlayer = false, StatusToUse is populated with localStatus value (not undefined)

    - Commented sync media provider
        - This code is not working as it supposed to and hindering video to play 
- Inside the Media Player Video
    - Changed the code to fix the issue : The issue was that when you play a video from video tab it will play, but when when you click on the video listed below it, it won’t play
    - This issue is still present in our live site.
- Updated the setToastError with new api setToast

PR #1070, opened 2024-03-11
fix: scroll on nft marketplace for mobile

### LINKED ISSUE #1059: Mobile: Buy NFT screen
Issues;
Buy NFT:
- Can’t scroll down 

Marketplace collection screen 
- Can’t scroll down. Layout issues

I believe this is a regression, at least scroll down was working.

sachin-into:
fixes #1059


omniwired (dismissed):

packages/components/nfts/NFTs.tsx
```diff
@@ -66,6 +66,7 @@ export const NFTs: React.FC<{
           )}
           minElemWidth={250}
           gap={layout.spacing_x2}
+          noFixedHeight={Platform.OS !== "web"}
```
n0izn0iz, line 69: this will load up to 10000 nfts, we can't do this here!

n0izn0iz (changes requested):

packages/components/nfts/NFTs.tsx
sachin-into: ok

packages/components/nfts/NFTs.tsx
```diff
@@ -53,21 +52,38 @@ export const NFTs: React.FC<{
...
+                data={info.item}
+                style={{ width: elemWidth }}
+              />
+            )}
+            minElemWidth={250}
+            gap={layout.spacing_x2}
+          />
+        ) : (
+          <FlatList
+            data={nfts}
```
sachin-into, line 72: Since header and other options are on other components. Just created FlatList for mobile

n0izn0iz: I think this is the same behavior, gridlist is already a flatlist, with noFixedHeight it's like the flatlist you added, did you check network request to see if it does not load everything?

sachin-into: Yeah, it's same.
I did check on item length and it was fetching next 100 item at a time.

omniwired: Thanks for confirming

PR #1075, opened 2024-03-11
feat(wallet): mainnet vs testnet network select

### LINKED ISSUE #1041: Mobile: Change network button
Values hardcoded 
Also misleading if is cosmo chain or if is chat related

omniwired:
image

fixes #1041


n0izn0iz:
please use networks package for networks list and management

sachin-into (dismissed):

packages/screens/Mini/Network/ChangeNetworkScreen.tsx
```diff
@@ -28,20 +33,22 @@ export default function ChangeNetworkScreen({
   navigation,
 }: ChangeNetworkScreenProps) {
   const goBackTo = () => navigation.replace("MiniSettings");
+  const selectedWallet = useSelectedNativeWallet();
 
   const addresses: ChangeNetworkType[] = [
```
n0izn0iz, line 38:
```ts
const enabledNetworks = useEnabledNetworks()
```
and
```suggestion
  const addresses: ChangeNetworkType[] = enabledNetworks.map(n => ({
    id: n.id,
    label: n.displayName,
    url: n.rpcEndpoint
  }))
```
you could add a filter if needed

n0izn0iz (changes requested):

packages/screens/Mini/Network/ChangeNetworkScreen.tsx
omniwired: thanks for your suggestion.
I tried to do it this way, but in this case doesn't make sense, because of this line.

https://github.com/TERITORI/teritori-dapp/blob/68f9e795d80541347a215c49e350ee678079486a/packages/hooks/useEnabledNetworks.ts#L12
The mini mode doesn't have an "enable testnet" setting.

The changes already in place are the toggle in mini mode at least.

It makes is more complex for no added benefit. (the mobile wallet only supports teritori at this point)

n0izn0iz: also I don't think we should use an intermediary network reprensentation (eg ChangeNetworkType)

n0izn0iz: don't blindly follow the figma, konolee doesn't know everything about our architecture and so some things might not match

n0izn0iz (dismissed):
I approve to allow for testnet mobile dev but I'd reeally like that we reuse/adapt the existing network management uis

packages/screens/Mini/Network/ChangeNetworkScreen.tsx
```diff
@@ -73,14 +66,35 @@ export default function ChangeNetworkScreen({
...
                 rightLabel: (
                   <View>
                     <Checkbox
-                      isChecked
-                      value={item?.label}
-                      onPress={() => {}}
+                      isChecked={selectedWallet?.networkId === item?.id}
+                      value={item?.displayName}
+                      onPress={() => {
+                        goBackTo();
```
hthieu1110, line 77: I think it'd better if we place goBackTo() at the end of other actions. This is kind of last action here. When replacing view, sometimes it stops all rendering and can cause some warning/errors because the code is trying to update inexist component.

packages/utils/wallet/getNativeSigner.ts
```diff
@@ -7,7 +7,7 @@ import { StoreWallet } from "@/utils/types/wallet";
 
 export const getNativeSigner = async (selectedWallet: StoreWallet) => {
   const network = mustGetCosmosNetwork(selectedWallet.networkId);
-
+  console.log("network", network);
```
hthieu1110, line 10: Remove this

n0izn0iz (approved):

PR #1079, opened 2024-03-14
feat(wallet): make the signer aware of multiple native wallets ids

omniwired:


packages/networks/signer.ts
```diff
@@ -21,6 +21,10 @@ import {
   keplrChainInfoFromNetworkInfo,
   mustGetCosmosNetwork,
 } from "@/networks/index";
+// eslint-disable-next-line no-restricted-imports
```
omniwired, line 24: @n0izn0iz 
I could relocate the slices, but I prefer the current location. Should I move it to override the linter?`

n0izn0iz: the slice is good in store package, would be better to pass the wallet slice to the function and use selector hook to get the data or something similar so the networks package doesn't depend on store

packages/utils/wallet/getNativeWallet.ts
```diff
@@ -13,6 +13,7 @@ export const getNativeWallet = (
   return (async () => {
     try {
       let mnemonic = await getMnemonic(index);
+      console.log("mnemonic", mnemonic);
```
n0izn0iz, line 16: remove debug?

packages/utils/wallet/getNativeWallet.ts
```diff
@@ -30,9 +31,11 @@ export const getNativeWallet = (
 
 export const getMnemonic = async (index: number = 0) => {
   try {
+    console.log("getting mnemonic", index);
```
n0izn0iz, line 34: remove debug?

n0izn0iz (changes requested):

omniwired:
> the slice is good in store package, would be better to pass the wallet slice to the function and use selector hook to get the data or something similar so the networks package doesn't depend on store

The problem with that approach is that I need to make every single contract aware of this function.
The original idea was to shim together and minimize the amount of changes.

n0izn0iz:
then maybe the native wallet should not use the redux store and instead use an isolated zustand store or a custom object with event emitters and a wrapping react context like other wallets? (I tend to think the zustand store is a good idea)

n0izn0iz:
multi-account is not in scope of the first release so maybe we can put this on ice?

Issue #1040, opened 2024-03-07
Mobile: Connected apps button


omniwired:
It doesn't do anything at the moment.

And maybe there is no need for it anyway, as this wallet can't be used in the web. 

There are no connected web apps, only the All-In-One experience.

PR #1097, opened 2024-03-15
fix(wallet): Connected apps button #1040

omniwired:
The wallet only work on the native app, not on the a webview.


n0izn0iz (approved):
LGTM

Issue #1086, opened 2024-03-14
Mobile: Normal mode Home Screen, picture keeps getting bigger


omniwired:
![image](https://github.com/TERITORI/teritori-dapp/assets/322761/6c9dfa4c-3f3f-4a23-9ad3-51146d772b51)

sujal-into:
Working on this issue

PR #1100, opened 2024-03-18
fix: home banner mobile layout

sujal-into:
issues: #1086

### DIFF
--- a/packages/components/hub/HubLanding.tsx
+++ b/packages/components/hub/HubLanding.tsx
@@ -1,3 +1,4 @@
+import { Link } from "@react-navigation/native";
 import React from "react";
 import { View, Linking } from "react-native";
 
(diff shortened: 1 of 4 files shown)


omniwired (approved):

packages/components/hub/HubLanding.tsx
```diff
@@ -33,12 +33,17 @@ export const HubLanding: React.FC = () => {
...
+    : "";
 
   return (
     <View style={{ alignItems: "center", width: "100%" }}>
       <View style={{ flex: 1 }}>
         {!!banner && (
-          <Link to={banner?.url || ""}>
+          <CustomPressable
+            onPress={() => navigation.navigate("FeedPostView", { id: postId })}
+          >
```
n0izn0iz, line 46: why are you changing this behavior? the banner url can be an external link

n0izn0iz (changes requested):

packages/components/hub/HubLanding.tsx
sujal-into: Currently the banner url is /feed/post/tori-bcd1253e-8b1a-4268-b3d8-59a1541d3fe2 (not external one). So I thought the only url we get is for feedpostview. If it can also be the external link  then, I will adjust that one. I used CustomPressable because the Link component previously used was causing the image to automatically grow bigger.

sujal-into: Fixed the problem with original Link component

n0izn0iz (approved):
LGTM

PR #1112, opened 2024-03-18
fix: hide connect ledger from  NativeWallet and add account screen

### LINKED ISSUE #1105: Mobile: Hide Ledger buttons
The reason is that we don't support it at the moment!

Here is an example but is not the only one.

image

sachin-into:
fixes #1105


omniwired (approved):
✨

n0izn0iz (approved):
LGTM

PR #1116, opened 2024-03-19
feat(wallet): TX detail screen

omniwired:


n0izn0iz (changes requested):
- the heuristics are not okay (eg `type.includes("send")`)
- we already have transaction details and preview components, check multisig transactions

Issue #1113, opened 2024-03-18
let's only expose seed export via mnemonic word for now


omniwired:
let's only expose seed export via mnemonic word for now

_Originally posted by @n0izn0iz in https://github.com/TERITORI/teritori-dapp/pull/1061#pullrequestreview-1942809620_

PR #1117, opened 2024-03-19
fix(wallet): private key import export is not supported

omniwired:
let's only expose seed export via mnemonic word for now #1113


clegirar (approved):
LGTM 👍

PR #1127, opened 2024-03-20
fix: share link url updated

ChiragPansuriya-iView:


omniwired (approved):

WaDadidou (approved):

n0izn0iz (changes requested):
like I said, there is a medium refactor needed, ShareButton should take the network-prefixed-postid as prop, using the selected network is not reliable, and will break when we will have cross-chain feed, also the `#Teritori` anchor is very weird

omniwired (approved):

n0izn0iz (changes requested):
See https://github.com/TERITORI/teritori-dapp/blob/main/CONTRIBUTING.md#multi-networks-multi-chains

PR #1130, opened 2024-03-21
feat: deep linking

sachin-into:


n0izn0iz (approved):
code LGTM

omniwired (dismissed):
😍

n0izn0iz (changes requested):
iOS build fails and PR has conflicts

app.config.js
```diff
@@ -37,6 +38,7 @@ const config = {
           "To optimize app launch times and resource utilization based on system boot information.",
         UIBackgroundModes: ["audio"],
       },
+      // associatedDomains: ["applinks:app.teritori.com"], FIXME: when adding this ios is not build but it is needed for deep linking
```
n0izn0iz, line 41: so deep linking doesn't work?

n0izn0iz (changes requested):

app.config.js
sachin-into: When it is enabled ios build failed. so, yeah it didn't work.

PR #1131, opened 2024-03-21
Fix/android UI fix

sujal-into:


omniwired (approved):

packages/screens/Mini/layout/MultiStepScreenContainer.tsx
```diff
@@ -24,13 +25,15 @@ export default function MultiStepScreenContainer({
   const navigation = useAppNavigation();
 
   const navigateBack = () => navigation.canGoBack() && navigation.goBack();
+  const safeAreaInsets = useSafeAreaInsets();
```
sujal-into, line 28: This is the fix for the issue: 1126

PR #1149, opened 2024-03-26
feat(wallet): disable multi wallets (Mobile)

omniwired:


omniwired:
https://github.com/TERITORI/teritori-dapp/pull/1079#issuecomment-2019223426

n0izn0iz (dismissed):
Code LGTM

packages/screens/Mini/Profile/ProfileScreen.tsx
```diff
@@ -88,6 +88,11 @@ export const ProfileScreen: ScreenFC<"MiniProfile"> = ({ navigation }) => {
     });
   };
 
+  // disable multiple wallet creation, remove to re enable multiple
```
hthieu1110, line 91: This comment is clear enough to me if I read this code here  but in some case if someone debug and follow the code, he can see that var is const then believe that its value will be immutable but not in this case.

so maybe we should create 2 consts :

const profileScreen = [<settings>, <dapps>] 
const profileForMultiWalletsScreen = [<add account>, ... profileScreen];

packages/screens/Wallet/Screens/NativeWallet.tsx
```diff
@@ -43,33 +45,47 @@ const NativeWallet: ScreenFC<"NativeWallet"> = () => {
...
+          }
           {nativeWallet && (
             <CustomButton
               title="Continue"
               style={{
                 marginVertical: layout.spacing_x2,
               }}
               onPress={(_, navigation) => {
-                navigation.navigate("MiniTabs", { screen: "MiniChats" });
+                if (appMode === "mini") {
```
hthieu1110, line 66: should we create enum AppMode instead of string ?

omniwired: an Enum doesn't make sense. The comparison is always against === "mini".
What I would make sense is to refactor this code everywhere and replace it with something like:
`useIsMiniMode()`

WaDadidou:
@hthieu1110 @n0izn0iz @omniwired 
What is the status of this PR ?
Is the issue still relevant ?

PR #1155, opened 2024-03-27
fix: upload video modal close icon bug fix

VishalRakholiya-iView:
Screenshot_1711543775

issue with close icon button, behave looks like static button, button there is onPress click event already present but no any action are there.

issue resolved.


n0izn0iz (changes requested):
don't comment code, remove it instead.

see https://methodpoet.com/commented-out-code/

n0izn0iz (approved):
LGTM

omniwired (approved):

PR #1160, opened 2024-03-29
fix: dropdown crop issue on account list of mobile

VishalRakholiya-iView:
Screenshot_1711692702
Screenshot_1711692711
Screenshot_1711692786
Screenshot_1711692790

The list of multiple account on mobile application side, there is crop issue on show dropdown,
when we have a one account, the first first account have facing crop issue, but we have multiple account then the last account of list have a crop issue, at the end of last item of list have a facing issue

issue resolved, Thanks.


omniwired (approved):

n0izn0iz (approved):
Code LGTM

PR #1161, opened 2024-03-29
fix(feed): multi-networkness, UI consistency and code dedup

n0izn0iz:
- disambiguate `post id` and `post local identifier` (local to network)
- properly use post network instead of selected network where it make sense
- use generic user components where possible
- simplify media player api
- better default video thumbnail
- misc cleanup


omniwired (changes requested):

omniwired (changes requested):
LGTM, one of the media assets should be optimized.

omniwired (dismissed):

go/cmd/teritori-dapp-backend/main.go
```diff
@@ -121,6 +121,7 @@ func main() {
 		Logger:    logger,
 		IndexerDB: indexerDB,
 		PinataJWT: *pinataJWT,
+		Networks:  netstore,
```
hthieu1110, line 124: Why we dont use Networks: network  :o

go/pkg/feed/service.go
```diff
@@ -25,6 +26,7 @@ type Config struct {
 	Logger    *zap.Logger
 	IndexerDB *gorm.DB
 	PinataJWT string
+	Networks  networks.NetworkStore
```
hthieu1110, line 29: if the value is NetworkStore then maybe naming attr name: netstore is less confusing ?

n0izn0iz: since it's acting like the frontend "networks" package, I though I could name it simpler, not a strong opinion

packages/components/socialFeed/SocialCard/SocialCardWrapper.tsx
```diff
@@ -55,21 +53,19 @@ export const SocialCardWrapper: FC<{
...
     refetch: refetchProposals,
-  } = useGetBanPostProposals(selectedNetworkId);
+  } = useGetBanPostProposals(networkId);
   const [isSendingProposal, setIsSendingProposal] = useState(false);
   const [isExecutingProposal, setIsExecutingProposal] = useState(false);
 
   const banPostProposalTitle = useMemo(() => {
-    return `Ban Post: ${post.identifier}`;
-  }, [post.identifier]);
+    return `Ban Post: ${post.localIdentifier}`;
```
hthieu1110, line 67: Just a note: I see that all other has been changed to : post.id, so maybe this is occasion to change this to => post.localId

packages/components/user/Username.tsx
```diff
@@ -15,39 +15,45 @@ export const Username: FC<{
...
-  const { primaryAlias } = useNSPrimaryAlias(userId);
-  const {
-    data: { allValidators },
-  } = useValidators(network?.id); // FIXME: this is ineficient
-  const validator = allValidators?.find((val) => val.address === userAddress);
-  if (isUserValidator(userId) && validator) {
+  namedColor?: ColorValue;
+  anonColor?: ColorValue;
+  prefix?: string;
+}> = memo(
```
hthieu1110, line 21: I see that we'd like to memoize component rendering, but if we use inline style which is object like ` {"attr": value} `,  for textStyle,  it will re-render this component every circle :D so `memo` here is useless, IMHO.

packages/hooks/feed/useFetchFeed.ts
```diff
@@ -30,25 +29,15 @@ const fetchTeritoriFeed = async (
...
-    return { list: [], totalCount: 0 } as PostsList;
-  }
+  const postsRequest: Partial<PostsRequest> = {
+    ...req,
+    offset: pageParam || 0,
+  };
+  const feedClient = mustGetFeedClient(selectedNetwork.id);
+  const response = await feedClient.Posts(postsRequest);
+  const list = response.posts.sort((a, b) => b.createdAt - a.createdAt);
+  console.log("fetchTeritoriFeed", list);
```
hthieu1110, line 39: should we remove this ?

packages/hooks/feed/useFetchFeed.ts
```diff
@@ -19,8 +18,8 @@ import { extractGnoJSONString } from "@/utils/gno";
 
 export type PostsList = {
   list: Post[];
-  totalCount: number;
-} | null;
+  totalCount: number | undefined;
```
hthieu1110, line 21: is this the same as :
```
export type PostsList = {
  list: Post[];
  totalCount?: number;
}
```

packages/components/user/Username.tsx
n0izn0iz: there is lot of uses where the style is constant

packages/components/socialFeed/SocialCard/SocialCardWrapper.tsx
n0izn0iz: I wanted to really disambiguate id and local id(entifier)

packages/hooks/feed/useFetchFeed.ts
n0izn0iz: nice catch

packages/hooks/feed/useFetchFeed.ts
n0izn0iz: not entirely, `totalCount: number | undefined;` requires that the field is defined, `totalCount?: number;` does not

PR #1163, opened 2024-03-30
fix: android crash due to weshnet; crashed due to go couldn't get net.InterfaceAddrs()

sakul-budhathoki:
Related issue:
https://github.com/golang/go/issues/40569

Fix deps:
go-multiaddr: https://github.com/TERITORI/go-multiaddr/pull/1
 
 
 Notes:
 - In local development, the app freezes after the weshnet starts (on the JS gets hang)
 - need to reload by pressing r then everything will work fine
 - need to check how it behaves on production


omniwired (approved):

n0izn0iz (approved):
Code LGTM

PR #1166, opened 2024-04-02
fix : android chat input UI, QR scanner when first use and added QR quietZone on web

sachin-into:


packages/components/modals/QRCodeScannerModal.native.tsx
```diff
@@ -62,15 +66,22 @@ export const QRCodeScannerModal = ({
...
+    >
       <View
         style={{
           height: height - 180,
           width: "100%",
         }}
       >
         <>
-          {device !== undefined && (
+          {device !== undefined && cameraPermissionStatus === "granted" && (
```
omniwired, line 84: 👀

omniwired (approved):
Code

n0izn0iz (approved):
Code LGTM, no regression on web

PR #1175, opened 2024-04-04
fix: loading state until wesh is connected

sachin-into:


weshd/go/shared/shared.go
```diff
@@ -168,7 +168,6 @@ func StartHTTPServer() {
 
 	if err := httpServer.ListenAndServe(); err != nil {
 		log.Println("weshnet: http listenAndServer err", err)
-		panic(errors.Wrap(err, "failed to start http server"))
```
omniwired, line 171: It's ok to continue with errors?

sachin-into: having it crashed app when disabling chat via toggle switch on android

omniwired (approved):

PR #1176, opened 2024-04-04
fix: Controlled component considered as uncontrolled because no value

WaDadidou:
Fixes this error :
`Warning: A component is changing an uncontrolled input to be controlled. This is likely caused by the value changing from undefined to a defined value, which should not happen. Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://reactjs.org/link/controlled-components`

(Btw `defaultValue` should be named `forcedValue` when I see its usage in `TextInputCustom` IMO)


n0izn0iz:
the problem seem to come from a bad usage of the component, I don't think we should add this hack

WaDadidou:
> the problem seem to come from a bad usage of the component, I don't think we should add this hack

Hm yes, have to set `value`

PR #1177, opened 2024-04-05
Fix/group chat

sachin-into:


packages/screens/Mini/Chat/components/ChatList.tsx
```diff
@@ -77,7 +77,15 @@ export const ChatList = () => {
         />
         <SpacerColumn size={3} />
         <FlatList
-          data={searchResults}
+          data={searchResults.filter((result) => {
+            if (
+              result?.type === "group" &&
+              (!result?.members || result?.members.filter(Boolean).length === 0) // Hiding conversation from left member of the group
```
omniwired, line 83: Wow, quite complex logic

weshd/go/shared/shared.go
```diff
@@ -168,7 +168,6 @@ func StartHTTPServer() {
 
 	if err := httpServer.ListenAndServe(); err != nil {
 		log.Println("weshnet: http listenAndServer err", err)
-		panic(errors.Wrap(err, "failed to start http server"))
```
omniwired, line 171: Isn't this still required or at least a good idea to have?

sachin-into: Yeah but having it crashed app when disabling chat via toggle switch on android

packages/screens/Mini/Chat/components/ChatList.tsx
sachin-into: when user leave group,  got members array with single 'undefined' element so filtered it out.

n0izn0iz: better fix at root of problem

sachin-into: ok

PR #1181, opened 2024-04-08
fix: avatar count on group chat

sachin-into:


packages/screens/Mini/Chat/components/ChatList.tsx
```diff
@@ -129,6 +129,15 @@ const SingleFriendChatList = ({ data, onPress }: SingleChatListType) => {
     onPress();
   };
 
+  const conversationMembers =
```
omniwired, line 132: This code may be good enough and simpler to understand, just like it is.

```
const conversationMembers = Array.isArray(conversation?.members)
  ? conversation.members.map((_, index) => getConversationAvatar(conversation, index))
  : [""];
  ```

I didn't test it

PR #1184, opened 2024-04-09
Fix/chat header space UI

VishalRakholiya-iView:
fix
old


VishalRakholiya-iView:
Screenshot_1712664021
button text issue

PR #1203, opened 2024-04-14
fix(multisig): refacto, force amino signing, disable keplr balance check and prevent modification by signer

n0izn0iz:


packages/hooks/multisig/useApproveTransaction.ts
```diff
@@ -61,30 +62,33 @@ export const useApproveTransaction = () => {
...
         };
 
-        const { bodyBytes, signatures } = await client.sign(
-          signerAddress,
-          tx.msgs,
-          tx.fee,
-          tx.memo,
-          signerData,
-        );
+        console.log("will sign", sd);
```
omniwired, line 79: maybe remove this one?

omniwired (dismissed):

PR #1208, opened 2024-04-16
fix: issue of navigate in MarketplaceScreen

ChiragPansuriya-iView:
Fixed the issue of navigation in MarketplaceScreen


packages/components/OmniLink.tsx
```diff
@@ -9,7 +9,7 @@ import {
 } from "react-native";
 
 interface OmniLinkToType {
-  screen: string | never;
+  screen: string | never | undefined;
```
omniwired, line 12: I feel this will bring problems.
It can also be 
> screen?

PR #1210, opened 2024-04-17
fix:(mobile) scroll issue fixed in profile page

VishalRakholiya-iView:


packages/screens/Mini/Profile/ProfileDetailScreen.tsx
```diff
@@ -170,7 +170,8 @@ export default function ProfileDetailScreen({
         style={{
           justifyContent: "space-between",
           backgroundColor: "rgba(0,0,0,0.8)",
-          height: "70%",
+          // height: "70%",
```
omniwired, line 173: why leave the commented-out code?

omniwired (changes requested):

PR #1222, opened 2024-04-29
feat: burn capital, landing page and multi burn WIP

omniwired:


n0izn0iz (changes requested):
it must be a `send`, not `transfer` message, otherwise the nft is sent without executing the "receive" code in contract and is "lost"

omniwired:
we might need to re enable some of the collections so they can be listed in the UI
Options:
- Revert some of these changes https://github.com/TERITORI/teritori-dapp/pull/712/files
- Find another solution; a different backend endpoint or using only contracts.

PR #1223, opened 2024-05-14
Tables refacto

WaDadidou:
Instead of removing columns on mobile, I made Static columns and Scrollables columns. The scrollable ones can be scolled horizontally.
Example on Magic Eden (check on mobile format): https://magiceden.io/


omniwired (approved):

omniwired:
I didn't test it on an actual device, did you have that in mind when doing this refacto @WaDadidou ?

n0izn0iz (changes requested):
regression on headers and rows alignment
Screenshot 2024-05-16 at 13 26 16

WaDadidou:
> I didn't test it on an actual device, did you have that in mind when doing this refacto @WaDadidou ?
Yes, it works on a real device in Web mode

WaDadidou:
> regression on headers and rows alignment Screenshot 2024-05-16 at 13 26 16

Ah yea, fixed

n0izn0iz (changes requested):
Screenshot 2024-05-18 at 22 18 34
weird hover behavior, the first row stays hovered and also the current hovered row stays hovered when you move out of the table

n0izn0iz:
check if https://www.npmjs.com/package/react-native-table-component supports the behavior you want to achieve, it will probably be more robust than rolling our own implem

WaDadidou:
> Screenshot 2024-05-18 at 22 18 34 weird hover behavior, the first row stays hovered and also the current hovered row stays hovered when you move out of the table

Yes, since it's now two columns, I shared a hover event between the two columns of a row, but there is a latency, the hover is not correctly sync

WaDadidou:
> check if https://www.npmjs.com/package/react-native-table-component supports the behavior you want to achieve, it will probably be more robust than rolling our own implem

Ah yea, I'll test

omniwired (approved):

WaDadidou:
UPDATE: I removed the split between scrollable header/row and static header/row => The whole table scrolls horizontally now.
I made reusable components to compose easelly a table with header and rows.
Better standardisation

WaDadidou:
> the height of rows changes in some places, also the font weight on firefox, is this intended?

Which places ?

n0izn0iz:
1) here the rows height an spacing have changed (ff and chrome), and the text of the usd value too (firefox only)
    your pr:
    Screenshot 2024-06-21 at 18 18 20
     master:
    Screenshot 2024-06-21 at 18 18 43
2) here the layout changed (both), the colors of activity types are gone (both) and font changed (both ff and chrome, but different ways)
    your pr (ff):
    Screenshot 2024-06-21 at 18 24 27
    master (ff):
    Screenshot 2024-06-21 at 18 22 03
    your pr (chrome):
    Screenshot 2024-06-21 at 18 25 10
    master (chrome):
    Screenshot 2024-06-21 at 18 25 33

WaDadidou:
> 1. here the rows height an spacing have changed (ff and chrome), and the text of the usd value too (firefox only)
>    your pr:
>    Screenshot 2024-06-21 at 18 18 20
>    master:
>    Screenshot 2024-06-21 at 18 18 43
> 2. here the layout changed (both), the colors of activity types are gone (both) and font changed (both ff and chrome, but different ways)
>    your pr (ff):
>    Screenshot 2024-06-21 at 18 24 27
>    master (ff):
>    Screenshot 2024-06-21 at 18 22 03
>    your pr (chrome):
>    Screenshot 2024-06-21 at 18 25 10
>    master (chrome):
>    Screenshot 2024-06-21 at 18 25 33

### 1. 
- On main, the rows height is wrong: 
image
It should be 50px, according to the Figma:
image
- I just increased the font size for USD values (11px to 13px). On ActivityTable (origin/main), the TORI value and the USD value have the same font size.


### 2.
- In ActivityTable, I set the same font size as CollectionsTable. It makes non sense to use two differents fonts or rows height.
Just did simplier an dmore harmonized.
===> But I fixed the colors of activity types

PR #1235, opened 2024-06-30
fix: refacto marketplace leaderboard table

WaDadidou:
==> Refacto `MarketplaceLeaderboardTable` using the new table components.
It allows to display all the columns and preserve the harmonization of our tables.

### Before:
image

### After:
image

### Horizontally scrollable on small screens:
image


packages/screens/Marketplace/MarketplaceScreen.tsx
```diff
@@ -113,6 +113,8 @@ export const MarketplaceScreen: ScreenFC<"Marketplace"> = () => {
           filterText={filterText}
         />
       </View>
+
+      <SpacerColumn size={16} />
```
hthieu1110, line 117: maybe we should use space_x2 instead of hardcoded value ?

hthieu1110 (approved):

packages/screens/Marketplace/MarketplaceScreen.tsx
WaDadidou: `size` value is already multiplied by 8, as `space_xX` value.
We could refactor this, but not in this PR.
Before `SpacerColumn`, there was a big margin, used to add vertical space under the "tables with no element at the bottom"

hthieu1110 (approved):

n0izn0iz (approved):
LGTM

PR #1239, opened 2024-07-06
feat: gno test 4 + deploy scripts

omarsy:


gno/Makefile
```diff
@@ -0,0 +1,80 @@
...
+		-gas-fee="1ugnot" \
+		-gas-wanted="50000000" \
+		-broadcast="true" \
+		-chainid="${CHAIN_ID}" \
+		-remote="${REMOTE}" \
+		-pkgdir="./p/dao_interfaces" \
+		-pkgpath="gno.land/p/demo/${BASE}/dao_interfaces" \
+		${KEY}
+
+all: add_utf16_pkg add_ujson_pkg add_flags_index_pkg add_dao_interfaces_pkg add_social_feeds_realm add_social_feed
```
n0izn0iz, line 80: relying on this for the order is not idiomatic make and will brake while using parallel make, it's better to properly use dependencies

I think we should use ts scripts for this

n0izn0iz: I will merge current makefile as it's already useful

PR #1243, opened 2024-07-20
fix: close map modal when user clicks 'skip'

WaDadidou:
image


WaDadidou:
Obsolete regarding the PR 1266

PR #1246, opened 2024-07-22
feat: map location backend

Villaquiranm:


go/cmd/gno_social_feed_indexer/clientql/client.go
```diff
@@ -0,0 +1,120 @@
...
+	posts, err := indexerql.GetPostTransactions(context.Background(), client.gqlClient, fromBlock)
+	if err != nil {
+		return err
+	}
+
+	for _, transaction := range posts.Transactions {
+		for _, post := range transaction.Messages {
+			data, ok := post.Value.(*indexerql.GetPostTransactionsTransactionsTransactionMessagesTransactionMessageValueMsgCall)
+			if !ok {
+				fmt.Println("not ok, skipping")
```
n0izn0iz, line 45: the policy on current indexers is to stop in unexpected cases

also use a zap logger instead of fmt please

n0izn0iz (changes requested):
partial

go/pkg/indexerQL/indexerQL.go
```diff
@@ -0,0 +1,458 @@
+// Code generated by github.com/Khan/genqlient, DO NOT EDIT.
+
+package indexerql
```
n0izn0iz, line 3: I would rename this package something more explicit like `gnoindexerql`

api/feed/v1/feed.proto
```diff
@@ -57,4 +75,6 @@ message PostsRequest {
 
 message PostsResponse {
     repeated Post posts = 1;
+    repeated AggregatedPost aggregatedPosts = 2;
+    bool is_aggregated = 3;
 }
```
n0izn0iz, line 80: check protobuf linter (it runs in `check-codegen` CI job)

go/cmd/gno_social_feed_indexer/clientql/client.go
```diff
@@ -0,0 +1,120 @@
...
+	err := json.Unmarshal([]byte(data.Args[3]), &metadata)
+	if err != nil {
+		return indexerdb.Post{}, err
+	}
+	metadataJSON, err := json.Marshal(metadata)
+	if err != nil {
+		return indexerdb.Post{}, err
+	}
+	post := indexerdb.Post{
+		Identifier:           uuid.Must(uuid.NewV4()).String(),
```
n0izn0iz, line 85: we should not use a generated ID but instead use something unique to this post that is consistent after replays

go/cmd/gno_social_feed_indexer/main.go
```diff
@@ -0,0 +1,89 @@
...
+	fs := flag.NewFlagSet("teritori-indexer", flag.ContinueOnError)
+	var (
+		dbHost       = fs.String("db-indexer-host", "", "host postgreSQL database")
+		dbPort       = fs.String("db-indexer-port", "", "port for postgreSQL database")
+		dbPass       = fs.String("postgres-password", "", "password for postgreSQL database")
+		dbName       = fs.String("database-name", "", "database name for postgreSQL")
+		dbUser       = fs.String("postgres-user", "", "username for postgreSQL")
+		networksFile = fs.String("networks-file", "networks.json", "path to networks config file")
+		networkID    = fs.String("gno-network-id", "devLocal", "network id to index")
+		txIndexerURL = fs.String("gno-tx-indexer-endpoint", "http://localhost:8546/graphql/query", "Tx indexer GraphQL endpoint")
```
n0izn0iz, line 29: you should get this from the network definition

n0izn0iz (changes requested):

go/cmd/gno_social_feed_indexer/clientql/client.go
Villaquiranm: Changed it for 
`Identifier:           fmt.Sprintf("%s-%s", data.Caller, transaction.Hash),`

go/pkg/indexerQL/indexerQL.go
Villaquiranm: Done :)

go/cmd/gno_social_feed_indexer/clientql/client.go
Villaquiranm: Added zap logger and for now keeping it as it was

random-poster/package.json
```diff
@@ -0,0 +1,18 @@
+{
```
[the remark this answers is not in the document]
Villaquiranm: Done and renamed to `socialFeedRandomPoster`

api/feed/v1/feed.proto
Villaquiranm: Now is okay for this check. I needed to commit some files from another services like weshnet weird

go/cmd/gno_social_feed_indexer/clientql/client.go
n0izn0iz: this is not unique, what if someone post twice in same tx (currently possible)

we should be able to get the post local id with events or the tx result instead I think

Villaquiranm: Okay I see What do you think of having
`TransactionHash-blockHeight-IndexOfPostInTransaction` ?
maybe just `TransactionHash-IndexOfPostInTransaction` would be enough ?

n0izn0iz: this can work, but it would need to contain network id also since the tx hash could be duplicated on other networks

n0izn0iz: I think the model is not good in the first place, the primary key should be a composite key of `network_id` and `identifier` to correctly be cross-chain

and the `identifier` should only be the on-chain post id

n0izn0iz: this is kinda out of scope, let me think about the best way to resolve this

go/cmd/gno_social_feed_indexer/main.go
n0izn0iz: network definition is a cross-language thing (we should probably transition to protobuf for this at some point btw)

I added the field in typescript and defined it for the gnodev network

changes are in https://github.com/TERITORI/teritori-dapp/pull/1246/commits/a43d98f0295bbe05ee365d9cdd092748c317f1b2 if you're curious

api/feed/v1/feed.proto
n0izn0iz: I think the extra changes are due to a yarn.lock version change probably by some IDE

PR #1247, opened 2024-07-23
fix: feed comment

omarsy:


packages/utils/gno.ts
```diff
@@ -40,20 +40,8 @@ export const adenaDoContract = async (
...
-  const deliverResults = blockResult.results.deliver_tx || [];
-  if (deliverResults.length <= index) {
-    throw new Error("tx result not found in block");
-  }
-  const err = deliverResults[index].ResponseBase.Error;
-  if (err) {
-    console.error(deliverResults[index]);
-    throw new Error(JSON.stringify(err));
-  }
+  await client.waitForTransaction(hash, height);
```
n0izn0iz, line 43: please keep the version that waits for the real results, otherwise we can't know if the transaction did succeed

n0izn0iz (changes requested):

packages/utils/gno.ts
n0izn0iz: I upgraded the patch, it should work

PR #1250, opened 2024-08-06
feat: use `r/demo/profile` for gno user info

### LINKED ISSUE #1244: feat: use new profile realm in upp / user data
use this realm: https://github.com/gnolang/gno/pull/1983 as upp/user data provider

hthieu1110:
Closes #1244 

Use Gno r/demo/profile for UPP page. Here are some results:

Screenshot 2024-08-06 at 23 27 06
Screenshot 2024-08-06 at 23 27 10
Screenshot 2024-08-06 at 23 27 17
Screenshot 2024-08-06 at 23 27 20


n0izn0iz (changes requested):
I think we shouldn't add a new hook and instead adapt the existing

- useNSPrimaryAlias (returns r/demo/users name)
- useNSUserInfo (returns relevant fields in r/demo/profiles)

without changing interfaces

n0izn0iz (changes requested):
we should try to keep changes minimal, so remove dev artifacts, this makes the changes more atomic, make reviews easier and prevents including unwanted changes

thus can you revert the changes in
- packages/components/images/AvatarWithFrame.tsx
- packages/screens/UserPublicProfile/components/UPPIntro.tsx

PR #1256, opened 2024-08-13
fix: install node modules before electron build to fix node caching

n0izn0iz:


n0izn0iz:
will be done in #1254

PR #1277, opened 2024-09-16
feat: use r/user for username and r/profile for profile metadata on gno network

hthieu1110:
Update UPP Form, supports networks: Gno, Teritori
- On Gno:
  - Support profile info without username
  - Support updating profile info, if nothing changed then profile do nothing when submit
  - Support adding username anytime (even after having profile info) but username is not changeable
  - Support checking availability, fee of username
  - Support atomic update

- On Teritori:
  - User have to register the username/tokenId along with profile info
  - When submit form, even without changes, we will create the transaction (teritori does not support atomic change)
  - Support checking availability, fee of username
  - No atomic update, we have to create a tx with full profile info when updating just 1 info

Here are some screenshots:
Screenshot 2024-09-16 at 3 06 43 PM
Screenshot 2024-09-16 at 3 06 58 PM
Screenshot 2024-09-16 at 3 08 25 PM
Screenshot 2024-09-16 at 3 08 40 PM


packages/hooks/useNSNameInfo.ts
```diff
@@ -10,18 +10,23 @@ import {
...
+import { ProfileData } from "@/utils/upp";
+
+export const GNO_CONTRACT_FIELD = {
+  // Standard fields
+  DISPLAY_NAME: "DisplayName",
+  BIO: "Bio",
+  AVATAR: "Avatar",
+
+  // Extra fields
+  BANNER: "Ext_Banner",
```
n0izn0iz, line 22: Not sure we need to prefix, this could actually become standard later and if we prefix now, we will have to support two fields

n0izn0iz (changes requested):
please add an e2e test so the CI can check that edit profile actually works

packages/screens/UserPublicProfile/components/modals/EditProfileModal.tsx
```diff
@@ -0,0 +1,174 @@
...
+
+        msgs.push({ type: "/vm.m_call", value: vmCall });
+      }
+
+      const height = await provider.getBlockNumber();
+      const txHash = await adenaDoContract(network.id, msgs, {
+        gasWanted: 2_000_000,
+      });
+      // Wait for tx done
+      await provider.waitForTransaction(txHash, height, 30 * 1000);
```
n0izn0iz, line 120: adenaDoContract is already waiting for the tx result

n0izn0iz (changes requested):

n0izn0iz (changes requested):
can you use the existing name field (like in current profile edit) and put everything in one modal instead of two steps?

n0izn0iz (changes requested):
it should support teritori also

hthieu1110:
> can you use the existing name field (like in current profile edit) and put everything in one modal instead of two steps?

I think separating in 2 steps make more sense: first step is mandatory, once you create the username then you have a profile and you can set the profile metadata. And further more, username cannot be changed and all other profile metadata can be changed so there are 2 different data types

n0izn0iz:
on gno you can fill profile with no username

packages/hooks/useNSNameInfo.ts
```diff
@@ -103,11 +117,13 @@ export const useNSNameInfo = (
...
   tokenId: string | null | undefined,
   enabled?: boolean,
+  // We can provide address for case where gno user has no tokenId/username but has profile info
+  gnoAddress?: string,
 ) => {
   const { data: nsInfo, ...other } = useQuery(
     nsNameInfoQueryKey(networkId, tokenId),
     async () => {
-      if (!tokenId) return null;
+      if (!tokenId && !gnoAddress) return null;
```
n0izn0iz: I don't link this change, we should avoid network-specific things in interfaces

gno uses the concept of `AddressOrName`, maybe we should accept this instead of `tokenId` here

n0izn0iz (changes requested):

n0izn0iz (changes requested):
please remove commented code

hthieu1110:
> please remove commented code

Done

PR #1278, opened 2024-09-17
feat: mark gifs as PostCategory.Picture

WaDadidou:
Posts with GIFs get now the `PostCategory` `Picture`, and are visible in Pics Feed
image
image

Users can make posts with only GIFs and no text, as with images files. So, it makes sense to treat these two case as same `PostCategory`

---
Also, I need that for the Map Feed, to correctly display the GIFs as `Picture` posts on the map. 
image
If not, the posts with GIFs will appear as `Normal` posts with empty content (Because the text is not mandatory)
image

### DIFF
--- a/packages/utils/feed/queries.ts
+++ b/packages/utils/feed/queries.ts
@@ -98,27 +98,22 @@ export const getPostCategory = ({
   message,
   gifs,
 }: NewPostFormValues): PostCategory => {
-  let category: PostCategory;
-  if (gifs?.length) {
-    category = PostCategory.Picture;
-  } else if (files?.length) {
-    if (files[0].fileType === "image") {
-      category = PostCategory.Picture;
-    } else if (files[0].fileType === "audio") {
-      category = PostCategory.Audio;
-    } else {
-      category = PostCategory.VideoNote;
+  if (gifs?.length) return PostCategory.Picture;
+  if (files?.length) {
+    switch (files[0].fileType) {
+      case "image":
+        return PostCategory.Picture;
+      case "audio":
+        return PostCategory.Audio;
+      default:
+        return PostCategory.VideoNote;
     }
-  } else if (title) {
-    category = PostCategory.Article;
-  } else if (message.startsWith("/question")) {
-    category = PostCategory.Question;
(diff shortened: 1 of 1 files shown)


n0izn0iz:
I think GIF is more a VideoNote ^^

it's low fps but still a video

MikaelVallenet (dismissed):

Villaquiranm (dismissed):

packages/utils/feed/queries.ts
```diff
@@ -96,9 +96,12 @@ export const getPostCategory = ({
   title,
   files,
   message,
+  gifs,
 }: NewPostFormValues): PostCategory => {
   let category: PostCategory;
-  if (files?.length) {
+  if (gifs?.length) {
```
omarsy, line 102: What do you think about this ?
```suggestion
  if (gifs?.length) {
    return PostCategory.Picture;
  }
  
  if (files?.length) {
    switch (files[0].fileType) {
      case "image":
        return PostCategory.Picture;
      case "audio":
        return PostCategory.Audio;
      default:
        return PostCategory.VideoNote;
    }
  }
  
  if (title) {
    return PostCategory.Article;
  }
  
  if (message.startsWith("/question")) {
    return PostCategory.Question;
  }
  
  if (message.startsWith("/generate")) {
    return PostCategory.BriefForStableDiffusion;
  }
  
  return PostCategory.Normal;
```

WaDadidou:
> I think GIF is more a VideoNote ^^
> 
> it's low fps but still a video

Would you see a GIF displayed in a player like this ?
image

Also, a GIF could be only one frame.

Also, users can add many GIFs and many images in one post. I display `Picture` posts on map in a small carousel.
I think it really makes sense to display GIFs like this.
image

packages/utils/feed/queries.ts
WaDadidou: It's out of scope since this PR just adds a condition in `getPostCategory` and doesn't refacto other conditions.
But it's better like this !

n0izn0iz:
well, maybe we shouldn like most other platform, differentiate betweeen short video that autoplays and long videos that need a click instead

but it's out of scope

PR #1280, opened 2024-09-18
Feat: Marketing contract

WaDadidou:
This contract allows to:
- Update and Get the Banner (home)
- Update and Get the Highlitghted News (home and top menu)
- Update and Get the Live Collections (launchpad and top menu)
- Update and Get the Upcoming Collections (launchpad)
- Update and Get the Highlitghted Collections (launchpad)
- Update the Admin

All the Update `#[msg(exec)]` can only be executed by the admin (Compares the `sender` address and the `admin_addr`)


cosmwasm-contracts/marketing/config.json
```diff
@@ -0,0 +1,4 @@
+{
+  "admin_adrr": "tori1yzjgaql23yxlwxxvmszssg23w3f8k6k2q75jss"
```
hthieu1110, line 2: admin_addr could be better

cosmwasm-contracts/marketing/src/contract.rs
```diff
@@ -0,0 +1,273 @@
...
+    if ctx.info.sender != config.admin_addr {
+      return Err(ContractError::Unauthorized);
+    }
+    // Replace all news with the new ones
+    self.news.clear(ctx.deps.storage);
+    for (index, news_item) in news.into_iter().enumerate() {
+        self.news.save(ctx.deps.storage, index as u64, &news_item)?;
+    }
+
+    Ok(Response::new().add_attributes(attributes))
```
hthieu1110, line 142: I think we attributes is not use in earlier code then you can write it here to make the code more readable, we dont have to traceback to see the value of attributes

cosmwasm-contracts/marketing/src/contract.rs
```diff
@@ -0,0 +1,273 @@
...
+    if ctx.info.sender != config.admin_addr {
+      return Err(ContractError::Unauthorized);
+    }
+    // Replace all live_collections with the new ones
+    self.live_collections.clear(ctx.deps.storage);
+    for (index, live_collection) in live_collections.into_iter().enumerate() {
+        self.live_collections.save(ctx.deps.storage, index as u64, &live_collection)?;
+    }
+
+    Ok(Response::new().add_attributes(attributes))
```
hthieu1110, line 164: Ok(Response::default()
            .add_attribute("action", "update_live_collections")
     )

make the code more readable. and seems that Response::default() add some needed default value to response also

cosmwasm-contracts/marketing/src/contract.rs
```diff
@@ -0,0 +1,273 @@
...
+    .news
+    .range(ctx.deps.storage, None, None, cosmwasm_std::Order::Ascending)
+    .map(|item| item.map(|(_, news_item)| news_item))
+    .collect::<StdResult<Vec<News>>>()?; 
+
+    Ok(news)
+  }
+
+  #[msg(query)]
+  pub fn get_live_collections(&self, ctx: QueryCtx) -> StdResult<Vec<MarketingCollectionPreview>> {
```
hthieu1110, line 240: Should we merge 3 queries: live, highlighted, upcoming => 1 query function and use a switch case to return the according data ? we can prevent the repeated code => maintain easier, contract size smaller,  easy to add more query with the same kind of data

cosmwasm-contracts/marketing/config.json
WaDadidou: The value or the key ?

cosmwasm-contracts/marketing/src/contract.rs
WaDadidou: Ok I see

cosmwasm-contracts/marketing/src/contract.rs
WaDadidou: After reflexion, handling "Which collections are shown in the Launchpad" doesn't make sense.
The collections shown are handled by the Marketplace, automatically, no need to choose.
I'll remove the collections stuff from the marketing contract

toorol: It's very nice that you are working!

PR #1285, opened 2024-09-25
feat: gno role manager contract

MikaelVallenet:


gno/p/role_manager/role_manager.gno
```diff
@@ -0,0 +1,406 @@
...
+// well known roles
+const (
+	ROLE_USER = "user"
+)
+
+// well known permissions
+const (
+	PERM_ALL         = "all"
+	PERM_ADD_ROLE    = "add:role"
+	PERM_REMOVE_ROLE = "remove:role"
```
thehowl, line 21: doesn't this make more sense like `role:add`? so that they also display better when ordering globally.

gno/p/role_manager/role_manager.gno
```diff
@@ -0,0 +1,406 @@
...
+)
+
+// well known roles
+const (
+	ROLE_USER = "user"
+)
+
+// well known permissions
+const (
+	PERM_ALL         = "all"
```
thehowl, line 19: for this one, I would suggest you to support a "glob" syntax instead; this way you can have things like `role:*`, or any other `area:*` to grant all permissions in a specific area.

thehowl (commented):
Took a quick look. First of all, please change the constants from ALL_CAPS to PascalCase; this is just for idiomatic Go.

Secondly, there are cases where I think maps or slices actually make sense here. Like Role.permissions. When you have small amounts of data, they can be more efficient; I suggest you play with `gno test -print-runtime-metrics`.

Finally, if you have a hierarchy of roles you also need a way to reorder them. So I'd suggest you leave this be for now; simply have a sudo-user `owner`, and then have roles modifiable by those with the correct meta-roles (`role:add`, etc.).

gno/p/role_manager/role_manager.gno
MikaelVallenet: Yes indeed, i think have something like resources:actions could be nice like
``role:read,write,all`` with a way to create resources & actions. i will add this in V2, thanks for review :pray:

MikaelVallenet:
> Took a quick look. First of all, please change the constants from ALL_CAPS to PascalCase; this is just for idiomatic Go.
> 
> Secondly, there are cases where I think maps or slices actually make sense here. Like Role.permissions. When you have small amounts of data, they can be more efficient; I suggest you play with `gno test -print-runtime-metrics`.
> 
> Finally, if you have a hierarchy of roles you also need a way to reorder them. So I'd suggest you leave this be for now; simply have a sudo-user `owner`, and then have roles modifiable by those with the correct meta-roles (`role:add`, etc.). (you can also go the full-length, and implement the hierarchy as you were)

Thanks for review, i think i will try to split the pkg in V2 to have the hierarchy_role_manager as an extension of the role_manager, so i can keep a very simple & straight forward version for the basic pkg. :rocket:

n0izn0iz:
superseded by #1293 feel free to reopen if needed

PR #1286, opened 2024-09-25
fix: change ipfs gateway

WaDadidou:
Before 
image

After
image


n0izn0iz:
I wish you kept the PR atomic but I'll merge

PR #1288, opened 2024-09-29
feat: Add gh verify agent

Villaquiranm:


n0izn0iz:
maybe we should move that to a new repo?

gno_github_agent/.env
```diff
@@ -0,0 +1,5 @@
+GNO_MNEMONIC=source bonus chronic canvas draft south burst lottery vacant surface solve popular case indicate oppose farm nothing bullet exhibit title speed wink action roast
```
hthieu1110, line 1: Is this normal that we keep this mnemonic here ?

gno_github_agent/clientql/client.go
```diff
@@ -0,0 +1,142 @@
...
+
+	return nil
+}
+
+func (client *IndexerQL) getLastTreatedBlock() (int, error) {
+	var verification db.Verification
+	err := client.db.Model(&db.Verification{}).Where("status = ?", string(db.VerificationStatusVerified)).Order("id desc").First(&verification).Error
+
+	if err != nil {
+		if errors.Is(err, gorm.ErrRecordNotFound) {
```
hthieu1110, line 71: I dont see the purpose if this error checking :o both cases return the same value

gno_github_agent/clientql/client.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+func (client *IndexerQL) dealWithVerification(event *gnoindexerql.GetValidationRequestsTransactionsTransactionResponseEventsGnoEvent, blockHeight int) error {
+	var handle string
+	var callerAddress string
+	for _, attr := range event.Attrs {
+		if attr.Key == "handle" {
+			handle = attr.Value
+		}
+		if attr.Key == "from" {
```
hthieu1110, line 87: we should use `else if` to prevent one more unused comparison

gno_github_agent/.env
Villaquiranm: fixed on f4a2a1f469fa31fc20f1534f07e68ae1d7408f89

gno_github_agent/clientql/client.go
Villaquiranm: fixed on f4a2a1f469fa31fc20f1534f07e68ae1d7408f89

gno_github_agent/clientql/client.go
Villaquiranm: fixed on f4a2a1f469fa31fc20f1534f07e68ae1d7408f89

Villaquiranm:
> maybe we should move that to a new repo? if we keep it here pls move it to `go/cmd/gno-ghverify-agent`

I think we should indeed move it to his own repository, I'll do as soon as this PR is approuved

PR #1292, opened 2024-10-03
fix: gas estimation when updating gno profile

hthieu1110:


n0izn0iz:
would be better if we used dynamic gas estimation via simulation but will merge

PR #1293, opened 2024-10-04
feat: gno RoleManager

MikaelVallenet:
This follow the first PR: #1285 
I kept the previous PR open since this one is "simpler" version, more focus on being easily usable.

----------------------------------------

# Role Manager Package

The `role_manager` package provides role-based permission management for communities, DAOs, or any application requiring a role system. 

## Features

- **Roles:** Define roles like `user` and `admin` with specific permissions.
- **Permission Management:** Roles can have custom permissions (e.g., `write:role`).
- **User Assignment:** Users can be assigned multiple roles, and roles can be granted or revoked by permissioned users.
- **Ownership Control:** Admin actions are controlled using an `ownable` pattern.

### Key Functions
- `CreateNewRole`
- `DeleteRole`
- `AddPermissionToRole`
- `RemovePermissionFromRole`
- `AddRoleToUser`
- `RemoveRoleFromUser`
- `HasPermission`
- `HasRole`

> :warning: PR description generated by AI from code files :robot:


n0izn0iz:
I think having `PrevRealm` calls embedded will be problematic, especially when working with DAOs, when a dao instantiate it's role manager, it's owner should be the DAO and not the caller of the DAO initialization routine

there should be a version that take addresses as input and we can maybe have a PrevRealm wrapper if needed

gno/p/role_manager/role_manager.gno
```diff
@@ -0,0 +1,187 @@
...
+		permissions: avl.NewTree(),
+		users:       avl.NewTree(),
+	}
+	adminRole.permissions.Set(PermissionAll, struct{}{})
+
+	rm.roles.Set(RoleUser, userRole)
+	rm.roles.Set(RoleAdmin, adminRole)
+
+	return rm
+}
```
n0izn0iz, line 56: I'm not a fan of forcing the creation of these roles

Can we have a way to instantiate a blank RoleManager with just the owner/admin?

gno/p/role_manager/role_manager.gno
```diff
@@ -0,0 +1,187 @@
...
+)
+
+type Role struct {
+	name        string
+	permissions *avl.Tree // permission -> struct{}
+	users       *avl.Tree // user addr -> struct{}
+}
+
+type RoleManager struct {
+	owner *ownable.Ownable
```
n0izn0iz, line 27: I think we should allow to drop ownership, forcing the existence of a superadmin is against web3 promises

also I think the concept of owner is redundant with the concept of admin in that case and we can remove owner altogether

gno/p/role_manager/role_manager.gno
```diff
@@ -0,0 +1,187 @@
...
+	user, ok := rm.users.Get(addr.String())
+	if !ok {
+		uRole := rm.mustGetRole(RoleUser)
+		tree := avl.NewTree()
+		tree.Set(RoleUser, uRole)
+		rm.users.Set(addr.String(), tree)
+		uRole.users.Set(addr.String(), struct{}{})
+		return tree
+	}
+	return user.(*avl.Tree)
```
n0izn0iz, line 186: mutating state in getters is misleading

n0izn0iz:
we could merge this and iterate in the main branch if you wish

gno/p/role_manager/role_manager.gno
```diff
@@ -0,0 +1,187 @@
...
+)
+
+const (
+	RoleUser  = "user"
+	RoleAdmin = "admin"
+)
+
+const (
+	PermissionAll       = "all"
+	PermissionWriteRole = "write:role"
```
n0izn0iz, line 17: you did not address morgan's comments I think https://github.com/TERITORI/teritori-dapp/pull/1285#discussion_r1787527009 and https://github.com/TERITORI/teritori-dapp/pull/1285#discussion_r1787527738

MikaelVallenet: resolve partially in: [aa249ec](https://github.com/TERITORI/teritori-dapp/pull/1293/commits/aa249ec4a581a01be7407b0490bca3f87c25c564)

gno/p/role_manager/role_manager.gno
MikaelVallenet: resolve [aa249ec](https://github.com/TERITORI/teritori-dapp/pull/1293/commits/aa249ec4a581a01be7407b0490bca3f87c25c564)

gno/p/role_manager/role_manager.gno
MikaelVallenet: An ownable allow the owner to drop or transfer ownership isn't ?
I think i have to add a cap to export the owner variable or create a getter to make it accessible and then the owner can use the properties of an ownable cf. [78120da](https://github.com/TERITORI/teritori-dapp/pull/1293/commits/78120da8e0ad3276d6024b3e28d1f42113de5700)

gno/p/role_manager/role_manager.gno
MikaelVallenet: let's talk about this, i can understand for admin role but i like the fact that everyone is a basic user so if you call a function hasRole on a new addr it will act as an user and not as "this user does not exist". In this way you can affect some rights or roles to user that never interact before with your realms or you can let the users role without any permissions if you don't want to do this.

MikaelVallenet:
> I think having `PrevRealm` calls embedded will be problematic, especially when working with DAOs, when a dao instantiate it's role manager, it's owner should be the DAO and not the caller of the DAO initialization routine
> 
> there should be a version that take addresses as input and we can maybe have a PrevRealm wrapper if needed

i dont understand how it could be secure to just ask for address as arguments ? or maybe i did not understand what you mean

n0izn0iz:
it's secure if the code feeding the addresses is secure:

in the dao example: if you want to handle the dao as an entity in the permission system then you would need a more complex wrapper than just `PrevRealm` everywhere, you would need a way to check the "dao as a collective" permissions, which is `CurrentRealm`

you can make a PrevRealm wrapper

anyway, I don't think we should spend too much time debating until we test this with real usecases

gno/p/role_manager/role_manager.gno
n0izn0iz: yep it can, yep it needs to be exposed I think

gno/p/role_manager/role_manager.gno
n0izn0iz: you still add roles in HasPermission, HasRole which is a getter

I also find weird that you add roles in DeleteRole and stuff like that

I think we should `initUser` only in `AddRoleToUser`

gno/p/role_manager/role_manager.gno
n0izn0iz: > In this way you can affect some rights or roles to user that never interact before with your realms
I think this is a bad pattern, it breaks least permissions principle

but I understand some use case, if there is something you only allow to members then want at some point allow to everyone, mmh..
I'm not sure we should support it. another principle of security system is to prevent bad usage (shooting itself in the foot)

n0izn0iz:
okay the real important requests

- `initUser` everywhere is not good
- expose owner (you only expose the address, not the object)

after that we can merge and debate other things later

gno/p/role_manager/role_manager.gno
```diff
@@ -191,19 +163,10 @@ func (rm *RoleManager) mustGetRole(roleName string) *Role {
...
-	uRole.users.Set(user.String(), struct{}{})
-	return tree
-}
-
-func (rm *RoleManager) getUser(addr std.Address) *avl.Tree {
+func (rm *RoleManager) mustGetUser(addr std.Address) *avl.Tree {
 	user, ok := rm.users.Get(addr.String())
 	if !ok {
-		return nil
+		panic("user does not have any roles")
```
n0izn0iz, line 175: I think we should not panic here, so we can just return false in `HasRole` and `HasPermission` if `userRoles == nil`
they will be hard to use otherwise I think

`RemoveRoleFromUser` could be noop too if user doesn't exists instead of panic

n0izn0iz (changes requested):

n0izn0iz:
can't wait to see that integrated into daos :)

PR #1310, opened 2024-10-17
fix(dao): proposals list invalidation

WaDadidou:
Actually, we have to reload the page, after a proposal action, to see the new status.
This fix refetch the proposals correctly

### Start
Let's click on Approve !
image

### Without the fix, only the vote status is refreshed, not the whole proposal. 
Here, we see the "Voted" gauge is still at "0"
image

### With the fix, the proposal is refresh
Here, we see the fresh proposal status with the votes gauge "1/1" and so the Execute action is available
image


MikaelVallenet:
LGTM

MikaelVallenet (approved):
LGTM

n0izn0iz (changes requested):
there is an invalidation mechanism and it was working previously, you shouldn't have to pass this `refetch` can you check the `invalidateQueries` call instead please

WaDadidou:
> there is an invalidation mechanism and it was working previously, you shouldn't have to pass this `refetch` can you check the `invalidateQueries` call instead please

This doesn't work here: https://github.com/TERITORI/teritori-dapp/blob/fix-refetch-proposal/packages/components/dao/ProposalActions.tsx#L118

n0izn0iz:
okay, can you try to fix this then?

WaDadidou:
> okay, can you try to fix this then?

I tried but I didn't succeed

WaDadidou:
Fixed properly

packages/hooks/dao/useDAOProposals.ts
```diff
@@ -51,7 +51,7 @@ export const useDAOProposals = (daoId: string | undefined) => {
   const { daoProposals: cosmWasmDAOProposals, ...cosmWasmOther } =
     useCosmWasmDAOProposals(daoId);
   const { data: gnoDAOProposals, ...gnoOther } = useQuery(
-    [daoProposalsQueryKey(daoId), NetworkKind.Gno],
+    [...daoProposalsQueryKey(daoId), NetworkKind.Gno],
```
n0izn0iz, line 54:
```suggestion
    daoProposalsQueryKey(daoId),
```
the network kind is already encoded in the `daoId` since the network prefix is unique to the network and the kind can be derived from network info

packages/hooks/dao/useDAOProposals.ts
```diff
@@ -133,7 +133,7 @@ const useCosmWasmDAOProposals = (daoId: string | undefined) => {
   const proposalModuleAddress = daoFirstProposalModule?.address;
 
   const { data, ...other } = useQuery(
-    [daoProposalsQueryKey(daoId), NetworkKind.Cosmos],
+    [...daoProposalsQueryKey(daoId), NetworkKind.Cosmos],
```
n0izn0iz, line 136:
```suggestion
    daoProposalsQueryKey(daoId),
```

n0izn0iz (changes requested):

packages/hooks/dao/useDAOProposals.ts
WaDadidou: done

packages/hooks/dao/useDAOProposals.ts
WaDadidou: done

PR #1311, opened 2024-10-17
Feat: Make hash available in addition to CID when using IPFS

WaDadidou:
### Before
`useIpfs` allows to get the CID of uploaded files

image
Our files types embed this CID as `url`

image

### After
`useIpfs` allows to get the CID and the hash (before parsing it to CID) of uploaded files

image
Our files types embed the CID as `url` and the hash as `hash`

image

### DIFF
--- a/packages/components/teritoriNameService/MediaPreview.tsx
+++ b/packages/components/teritoriNameService/MediaPreview.tsx
@@ -68,11 +68,11 @@ export const MediaPreview: React.FC<{
       return;
     }
     if (documentPickerResult.output) {
-      const fileIpfsHash = await pinataPinFileToIPFS({
+      const pinFileToIPFSResult = await pinataPinFileToIPFS({
         pinataJWTKey,
         file: { file: documentPickerResult.output[0] },
       } as PinataFileProps);
-      callback(`ipfs://${fileIpfsHash}`);
+      callback(`ipfs://${pinFileToIPFSResult?.ipfsCid || ""}`);
     }
   };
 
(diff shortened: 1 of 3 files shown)


MikaelVallenet (approved):
LGTM

n0izn0iz:
what is the purpose of this? I think you are putting twice the same data, this is costing gas for no benefit form what I understand

n0izn0iz (changes requested):
requesting change to prevent merging before clarification

WaDadidou:
> requesting change to prevent merging before clarification

I send a hash and not a CID to UploadMetadatas:
https://github.com/TERITORI/teritori-dapp/blob/feat/cosmwasm-launchpad/packages/hooks/launchpad/useCompleteCollection.ts#L116

Because, in the UploadMetadatas function, we VerifyPinned:
https://github.com/TERITORI/teritori-dapp/blob/feat/cosmwasm-launchpad/go/pkg/launchpad/service.go#L78

My modification is stupid, I might as well adapt the service rather than change useIpfs, or maybe retrieve the hash from the CID

n0izn0iz:
closing, like discussed in sync:

in all cases: we shouldn't modify this basic type that is used to publish data on-chain
also: a cid is a hash
and: VerifyPinned is actually parsing a CID from a string

f you are passing the metadata.image field, which should be an ipfs uri, you need to strip the `ipfs://` scheme, maybe that was the actual problem 

also be careful to always store ipfs uris (`ipfs://<cid>`) on-chain and not raw CIDs

WaDadidou:
Bad files types modifications !

PR #1314, opened 2024-10-18
feat(file-uploader): move files + add `required` prop + allow to clear from parent

WaDadidou:
### Better arborescence
image

### Make `<Label isRequired` works
I've just added a `required` prop to `FileUploader`. We consider `FileUploader` as an "input". It has a Label and it is used in forms

image

---
###  Allow to clear the `input type='file'` in `FileUploaderSmall` from a parent using a ref
Not used yet, but used in Launchpad PR (https://github.com/TERITORI/teritori-dapp/pull/1024).
We can remove that from this atomic PR if you want

image
image

Usage

image

The point with that is that we can clear the file visually, that's ok (React state). BUT the file is still in the `input type='file'`.
So I've added this `resetFiles` cleaner.


MikaelVallenet (approved):
LGTM

PR #1317, opened 2024-10-21
feat: add useDAOProposalById

WaDadidou:
New hook: https://github.com/TERITORI/teritori-dapp/blob/feat-use-dao-proposal-by-id/packages/hooks/dao/useDAOProposalById.ts#L23

It uses existing backend stuff `listProposals` (Cosmos) and `getProposalJSON` (Gno)

Pretty the same as `useDAOProposals`, but for "proposal by id": https://github.com/TERITORI/teritori-dapp/blob/feat-use-dao-proposal-by-id/packages/hooks/dao/useDAOProposals.ts#L49


n0izn0iz (changes requested):
I think we should extract the conversion code instead of copying it

eg: create `gnoToAppProposal` and `cosmwasmToAppProposal` and use it in both "proposals" and "proposalById" hooks

WaDadidou:
> I think we should extract the conversion code instead of copying it
> 
> eg: create `gnoToAppProposal` and `cosmwasmToAppProposal` and use it in both "proposals" and "proposalById" hooks

Done

PR #1320, opened 2024-10-30
feat: merge cw address list contract

WaDadidou:


rust/cw-contracts/cw-address-list/Makefile
```diff
@@ -0,0 +1,54 @@
+ADMIN_TESTNET=testnet-adm
```
n0izn0iz, line 1: we should remove this makefile and use scripts that are common to all contracts IMO

n0izn0iz (changes requested):

rust/cw-contracts/cw-address-list/Makefile
WaDadidou: Done

packages/networks/types.ts
```diff
@@ -72,6 +72,8 @@ export type CosmosNetworkInfo = NetworkInfoBase & {
   daoVotingCw4CodeId?: number;
   cwAdminFactoryContractAddress?: string;
   coreDAOAddress?: string;
+  cwAddressListContractAddress?: string;
+  cwAddressListCodeId?: number;
```
n0izn0iz, line 76: please add a NetworkFeature if you need this in this PR

n0izn0iz (changes requested):

packages/scripts/integration-testing/cwAddresslistTest.ts
```diff
@@ -0,0 +1,33 @@
+import { program } from "commander";
+import fs from "fs/promises";
+
+import { deployCwAddressList } from "./../network-setup/deployCwAddressList";
```
n0izn0iz, line 4: using this import will also trigger the `main` func in `deployCwAddressList`

n0izn0iz (changes requested):

packages/scripts/integration-testing/cwAddresslistTest.ts
WaDadidou: Removed. 
We can test using deployCwAddressList with localnet

packages/networks/types.ts
WaDadidou: Done

packages/networks/features.ts
```diff
@@ -17,8 +17,19 @@ export enum NetworkFeature {
...
+  CosmWasmAddressList = "CosmWasmAddressList",
 }
 
+// CosmWasm Address List
+
+const zodCCosmWasmAddressList = z.object({
+  type: z.literal(NetworkFeature.CosmWasmAddressList),
+  cwAddressListContractAddress: z.string(),
+  codeId: z.number(),
+});
```
n0izn0iz, line 29: this feature should probably be called `NFTMarketplaceFeature`
there is no "address list feature", the address list is a tool that is used as the whitelist of the marketplace feature

we can move the other fields into it later

n0izn0iz (changes requested):

packages/networks/features.ts
WaDadidou: Wa sorry I'm dumb today

PR #1328, opened 2024-11-03
feat: handle NormalToast warning

WaDadidou:


WaDadidou:
This is usefull for NFT Launchpad front.
And it makes sense to have that.
And we have a warning MiniToast

packages/components/WalletProviderIcon.tsx
```diff
@@ -23,7 +24,14 @@ export const WalletProviderIcon: React.FC<{
     case WalletProvider.Adena:
       return <SVG width={size} height={size} source={adenaSVG} />;
     case WalletProvider.Gnotest:
-      return <SVG width={size} height={size} source={warningSVG} />;
+      return (
+        <SVG
+          width={size}
+          height={size}
+          source={warningSVG}
+          color={errorColor}
```
MikaelVallenet, line 32: should you use color & type: warning ? else do we use the warning type you add somewhere ?

WaDadidou: No.
I just set `color` here, because:
- warning.svg has now a dynamic `#color` instead of a static red
- So, when warning.svg is used, we need to set a `color` to `<SVG/>`.

I just needed to have a warning.svg with a dynamic `color`.

Now, I'll use this warning.svg with a `warningColor` in the NFT Launchpad PR.

WaDadidou: It's a common refacto: Instead of having many similar SVG with different colors, we can just have one SVG with a dynamic color prop.

We have a TON of SVGs in the assets folder

MikaelVallenet: Got it, LGTM :heart:

MikaelVallenet (approved):

PR #1334, opened 2024-11-03
feat(launchpad): add pluralize lib

WaDadidou:
Add the lib `pluralize` and `@types/pluralize` to handle plural strings.
And remove `pluralOrNot` function


packages/utils/text.tsx
```diff
@@ -34,4 +34,10 @@ export const replaceBetweenString = (
   `${origin.substring(0, startIndex)}${insertion}${origin.substring(endIndex)}`;
 
 export const pluralOrNot = (word: string, quantity: number) =>
```
hthieu1110, line 36: I think for most of cases , this is fine if we don't need the perfect solution :D 
But if we are perfectionist, we have to handle many edge cases like words finished by s (like gas) or ch ...

hthieu1110 (approved):
LGTM

MikaelVallenet (dismissed):

packages/utils/text.tsx
WaDadidou: Yes, I added the lib `pluralize` => https://classic.yarnpkg.com/en/package/pluralize

Villaquiranm (approved):
LGTM :)

n0izn0iz (approved):

MikaelVallenet (approved):

PR #1336, opened 2024-11-05
feat(gno/dao): create a proposal to post on social feed through dao

MikaelVallenet:


packages/utils/strings.ts
```diff
@@ -4,3 +4,7 @@ export const trimRight = (str: string, pattern: string) => {
   }
   return str;
 };
+
+export const capitalizeFirstLetter = (val: string) => {
+  return val.charAt(0).toUpperCase() + val.slice(1);
+};
```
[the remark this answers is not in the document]
MikaelVallenet: [9b2d5ba](https://github.com/TERITORI/teritori-dapp/pull/1336/commits/9b2d5ba6e510f10b6286f37936d810ece800c931)

packages/components/socialFeed/NewsFeed/NewsFeedInput.tsx
```diff
@@ -141,7 +141,7 @@ export const NewsFeedInput = React.forwardRef<
     const selectedNetwork = useSelectedNetworkInfo();
     const selectedNetworkId = selectedNetwork?.id || "teritori";
     const selectedWallet = useSelectedWallet();
-    const userId = getUserId(selectedNetworkId, selectedWallet?.address);
+    const userId = daoId ? daoId : selectedWallet?.userId;
```
n0izn0iz, line 144:
```suggestion
    const posterId = daoId ? daoId : selectedWallet?.userId;
```
let's make this ultra explicit

packages/hooks/feed/useFeedPosting.ts
```diff
@@ -38,7 +43,9 @@ export const useFeedPosting = (
   const [step, setStep] = useState(
     feedPostingStep(FeedPostingStepId.UNDEFINED),
   );
-  const { balances } = useBalances(networkId, userAddress);
+
+  const selectedWallet = useSelectedWallet();
+  const { balances } = useBalances(networkId, selectedWallet?.address);
```
n0izn0iz, line 48: are you sure about this change?

we should measure the poster (DAO in this case) funds for post fees and not the proposer funds

n0izn0iz (changes requested):

packages/hooks/feed/useFeedPosting.ts
MikaelVallenet: > are you sure about this change?
> 
> we should measure the poster (DAO in this case) funds for post fees and not the proposer funds

Hmm, i think i did not understood this at time i modified it.
Thought it was to ensure user have money to pay the proposal, but now i understand it was checking if the dao has the fee to pay the execution of proposal.

The thing is when the error was trigger a modal showing the balance of the user popped-up instead of showing the balance of the DAO, i will fix this and revert what you highlighted to ensure DAO have enough money!

Thanks for catching this :pray:

packages/components/socialFeed/NewsFeed/NewsFeedInput.tsx
MikaelVallenet: [ec11c14](https://github.com/TERITORI/teritori-dapp/pull/1336/commits/ec11c14bd8775973edf5da3029e04310db2f9783)

packages/hooks/feed/useFeedPosting.ts
```diff
@@ -74,34 +81,74 @@ export const useFeedPosting = (
...
+            func: "ProposeJSON",
+            args: ["0", JSON.stringify(propReq)],
+          };
+          const txHash = await adenaDoContract(
+            network.id,
+            [{ type: AdenaDoContractMessageType.CALL, value: vmCall }],
+            { gasWanted: 20_000_000 },
+          );
+          const provider = new GnoJSONRPCProvider(network.endpoint);
+          await provider.waitForTransaction(txHash);
```
n0izn0iz, line 119:
```suggestion
          await adenaDoContract(
            network.id,
            [{ type: AdenaDoContractMessageType.CALL, value: vmCall }],
            { gasWanted: 20_000_000 },
          );
```

`adenaDoContract` already waits on the tx

packages/components/modals/NotEnoughFundModal.tsx
```diff
@@ -28,13 +28,14 @@ export const NotEnoughFundsModal: FC<{
...
   onClose?: () => void;
-}> = ({ label = "Not enough funds", cost, visible, onClose }) => {
+}> = ({ label = "Not enough funds", cost, visible, onClose, address }) => {
   const selectedWallet = useSelectedWallet();
   const selectedNetwork = useSelectedNetworkInfo();
   const { balances } = useBalances(
     selectedNetwork?.id,
-    selectedWallet?.address,
+    address || selectedWallet?.address,
   );
```
n0izn0iz, line 39: we should pass an user id, not just an address, we don't know if the selectedNetwork matches this address' network

I would recommend to remove
```
const selectedWallet = useSelectedWallet();
const selectedNetwork = useSelectedNetworkInfo();
```
and instead take and `user id` or `network id + address` as props

see: https://github.com/TERITORI/teritori-dapp/blob/main/CONTRIBUTING.md#multi-networks-multi-chains

packages/components/socialFeed/NewsFeed/NewsFeedInput.tsx
```diff
@@ -141,7 +141,8 @@ export const NewsFeedInput = React.forwardRef<
     const selectedNetwork = useSelectedNetworkInfo();
     const selectedNetworkId = selectedNetwork?.id || "teritori";
     const selectedWallet = useSelectedWallet();
-    const userId = getUserId(selectedNetworkId, selectedWallet?.address);
+    const posterId = daoId ? daoId : selectedWallet?.userId;
```
n0izn0iz, line 144: nit, this can be simplified as

```suggestion
    const posterId = daoId || selectedWallet?.userId;
```

packages/components/socialFeed/NewsFeed/NewsFeedInput.tsx
```diff
@@ -141,7 +141,8 @@ export const NewsFeedInput = React.forwardRef<
...
+    const posterAddress = parseUserId(posterId)[1];
```
n0izn0iz, line 145: nit, the usual pattern to access hooks return is destructuring (hooks in official react docs are used like this)

```suggestion
    const [, posterAddress] = parseUserId(posterId);
```

n0izn0iz (changes requested):

packages/components/modals/NotEnoughFundModal.tsx
n0izn0iz: we could also just skip checking the funds for the DAO case since the funds can change between proposal creation and execution, it should only be a warning anyway

MikaelVallenet: I propose i do this in the next PR to keep small PR
What i can do is a warning at the moment of execution, for me it does not make sense to warn or block this at the proposal creation but should be at the execution, what do you think ?

Like warning: DAO does not have enough funds, execution is likely to fail. Do you still want to execute it ?

MikaelVallenet: [53793c3](https://github.com/TERITORI/teritori-dapp/pull/1336/commits/53793c3f32338cbb5eec4614aa486875066f4419)

packages/components/socialFeed/NewsFeed/NewsFeedInput.tsx
MikaelVallenet: i don't use this line anymore since i parse it in the modal
but i note this to avoid repeating this pattern in next PRs

packages/components/modals/NotEnoughFundModal.tsx
n0izn0iz: > for me it does not make sense to warn or block this at the proposal creation but should be at the execution, what do you think 
yes I agree sorry if it wasn't clear

n0izn0iz (approved):

WaDadidou (changes requested):

WaDadidou:
It's good, the balances are well used and the Gno DAO stuff seems to be well implemented on front

packages/components/socialFeed/NewsFeed/NewsFeedInput.tsx
```diff
@@ -141,7 +141,7 @@ export const NewsFeedInput = React.forwardRef<
     const selectedNetwork = useSelectedNetworkInfo();
     const selectedNetworkId = selectedNetwork?.id || "teritori";
     const selectedWallet = useSelectedWallet();
-    const userId = getUserId(selectedNetworkId, selectedWallet?.address);
+    const posterId = daoId || selectedWallet?.userId;
```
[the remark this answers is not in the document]
n0izn0iz: you're right we already use the term author instead of poster in other places my bad nice catch

n0izn0iz (approved):

packages/components/socialFeed/NewsFeed/NewsFeedInput.tsx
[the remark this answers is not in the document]
MikaelVallenet: [30f69f6](https://github.com/TERITORI/teritori-dapp/pull/1336/commits/30f69f6be3036b5e63c8b03edaf532dedf5eb1af)

WaDadidou (approved):
LGTM

PR #1340, opened 2024-11-06
fix: prevent sending NaN from DateTimeInput

WaDadidou:
Just check for `"Invalid date"` before triggering `onChange`

The issue appears when I use a form (with `react-hook-form`) and I want to keep a `DateTimeInput` empty.
The form has an error because the value of the` DateTimeInput` is `NaN`, but should be `undefined`


packages/components/inputs/DateTimeInput.tsx
```diff
@@ -101,7 +101,7 @@ export const DateTimeInput: FC<{
   };
 
   useEffect(() => {
-    if (error || !date) return;
+    if (error || !date || date.toString() === "Invalid Date") return;
```
MikaelVallenet, line 104: error should trigger the condition if date is invalid isn't ? or is it possible to does not have error but invalid date ?

WaDadidou: I'm not sure I understand your comment.

`"Invalid Date"` appears if the date can't be parsed (because the input is empty, or because other unplanned reasons).
We to `onChange` sends a valid date.
So I added the `"Invalid Date"` case to prevent `onChange` sending a broken date

MikaelVallenet: i mean if the date is equal to "Invalid Date", the error value should be filled and trigger the condition isn't ?

```js
  const error = useMemo(() => {
    if (!isDirty) return;
    if (!day && !month && !year && !hours && !minutes) {
      if (required) return DEFAULT_FORM_ERRORS.required;
      return;
    }
    if (
      date.toString() === "Invalid Date" ||
      parseInt(year, 10) === 0 ||
      parseInt(month, 10) === 0 ||
      parseInt(day, 10) === 0 ||
      year.length < 4 ||
      parseInt(month, 10) > 12 ||
      parseInt(month, 10) < 1 ||
      parseInt(day, 10) < 1 ||
      parseInt(day, 10) > 31
    ) {
      return DEFAULT_FORM_ERRORS.datetime;
    }
  }, [year, month, day, hours, minutes, date, isDirty, required]);
  ```

in this i understand that if date is equal to "Invalid Date" the error is equal to ``DEFAULT_FORM_ERRORS.datetime`` isn't ?

WaDadidou: We don't want to display an error if the field is unchanged, that's why I control `isDirty`.
So yes, it's possible to not have an error but invalid date (To anwser that: https://github.com/TERITORI/teritori-dapp/pull/1340#discussion_r1832233750)
Maybe I can force `date` to get `null` here: https://github.com/TERITORI/teritori-dapp/blob/1249427523eca01c44e400fb06af961f8cdae5ea/packages/components/inputs/DateTimeInput.tsx#L64
But it adds more complexity, and it's the same trick: `date` has a wrong value if the input is empty, and we don't want to send this value or show an error if the input is not dirty.

Maybe the whole logix needs a refacto and can be simpler, don'y hesitate to make a refacto in another PR

MikaelVallenet (approved):

n0izn0iz (approved):

PR #1341, opened 2024-11-07
feat(gno/dao): register dao conf to profile realm

MikaelVallenet:
depends on #1336 
add a link between dao deployment and profile realm and set name, bio, avatar to name, description, image at dao creation


n0izn0iz (approved):
LGTM

omarsy (approved):

PR #1354, opened 2024-11-15
feat(multisig): Add feedbacks to creation form

WaDadidou:
This PR add error feedbacks to the Multisig creation form. I added 2 try catch
It also add a loader to the `PrimaryButton`.

## Existing errors

image

image

image

## New errors

image

image

### DIFF
--- a/packages/screens/Multisig/MultisigCreateScreen.tsx
+++ b/packages/screens/Multisig/MultisigCreateScreen.tsx
@@ -1,8 +1,10 @@
 import { createMultisigThresholdPubkey } from "@cosmjs/amino";
-import React, { useEffect, useMemo, useState } from "react";
+import { zodResolver } from "@hookform/resolvers/zod";
+import React, { Fragment, useEffect, useMemo, useState } from "react";
 import { useForm } from "react-hook-form";
 import { Pressable, ScrollView, View } from "react-native";
 import { ActivityIndicator } from "react-native-paper";
+import { z } from "zod";
 
 import { MultisigSection } from "./components/MultisigSection";
 import trashSVG from "../../../assets/icons/trash.svg";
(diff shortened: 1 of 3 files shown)


hthieu1110 (commented):
Overall LGTM

n0izn0iz (changes requested):
This introduces a bug when you don't change the `Address #1` that is automatically injected, my guess it does not fetch the public key of this address

Screenshot 2024-11-15 at 08 45 20

(does not happen on master)

WaDadidou:
An issue:
When the form is valid, the PrimaryButton is still disabled, so the user has to manually modify an input to enable the button

WaDadidou:
> This introduces a bug when you don't change the `Address #1` that is automatically injected, my guess it does not fetch the public key of this address
> 
> Screenshot 2024-11-15 at 08 45 20
> (does not happen on master)

Fixed

packages/screens/Multisig/MultisigCreateScreen.tsx
```diff
@@ -43,27 +48,54 @@ import {
 } from "@/utils/style/fonts";
 import { layout } from "@/utils/style/layout";
 
-type CreateMultisigWalletFormType = {
-  addresses: { address: string }[];
-  signatureRequired: string;
-  maxSignature: string;
-  name: string;
-};
+const ZodCreateMultisigWalletFormType = z.object({
```
n0izn0iz, line 51:
```suggestion
const zodCreateMultisigWalletFormType = z.object({
```
we use capitalization only for types (in the typescript sense) and this is an object (a type schema)

n0izn0iz (changes requested):

packages/utils/errors.ts
WaDadidou, line 1: Btw I changed this error to "Required" because the default react-hook-form error for required fields is "Required", so it harmonizes

n0izn0iz (changes requested):
it's very weird that we have a different behavior before and after the first submit attempt IMO

packages/screens/Multisig/MultisigCreateScreen.tsx
```diff
@@ -118,39 +120,73 @@ export const MultisigCreateScreen = () => {
...
+      });
+    } catch (err) {
+      console.error("Failed to create multisig", err);
+      const message = err instanceof Error ? err.message : "";
+      setToast({
+        type: "error",
+        mode: "normal",
+        title: "Failed to create multisig",
+        message,
+      });
```
n0izn0iz, line 143: why do you add this? there is already a `wrapWithFeedback` around the submit call

packages/screens/Multisig/MultisigCreateScreen.tsx
```diff
@@ -118,39 +120,73 @@ export const MultisigCreateScreen = () => {
...
 
-    navigation.navigate("MultisigWalletDashboard", {
-      id: getUserId(selectedNetwork?.id, res.multisigAddress),
-    });
+      navigation.navigate("MultisigWalletDashboard", {
+        id: getUserId(selectedNetwork?.id, res.multisigAddress),
+      });
+    } catch (err) {
+      console.error("Failed to create multisig", err);
+      const message = err instanceof Error ? err.message : "";
```
hthieu1110, line 137: in case where err is not Error we can try to stringify the err (e.g: `${err}`) instead of sending ""

packages/screens/Multisig/MultisigCreateScreen.tsx
WaDadidou: No more zod usage. Outdated

packages/utils/errors.ts
WaDadidou: I revert that. Outdated

packages/screens/Multisig/MultisigCreateScreen.tsx
WaDadidou: https://github.com/TERITORI/teritori-dapp/pull/1354/commits/7666d919d204c3e6e70948b541ab6f42a76f0303

packages/screens/Multisig/MultisigCreateScreen.tsx
WaDadidou: I forgot `wrapWithFeedback`.
https://github.com/TERITORI/teritori-dapp/pull/1354/commits/7666d919d204c3e6e70948b541ab6f42a76f0303

packages/screens/Multisig/MultisigCreateScreen.tsx
```diff
@@ -118,39 +119,65 @@ export const MultisigCreateScreen = () => {
...
+      });
+    } catch (err) {
+      throw new Error(`Failed to create multisig: ${err}`);
+    } finally {
+      setLoading(false);
+    }
   };
 
-  const onAddressChange = async (index: number, value: string) => {
+  const onValidate = async (index: number, value: string) => {
```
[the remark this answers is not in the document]
WaDadidou: https://github.com/TERITORI/teritori-dapp/pull/1354/commits/e30148a77a4f86e42b42e2be703b6406de940096

n0izn0iz (approved):
LGTM

hthieu1110 (approved):
LGTM

PR #1372, opened 2024-11-16
fix(feed): Allow to focus Article Editor by clicking the placeholder

WaDadidou:
Actually, when the user clicks on "Type message here", it does nothing.
It's just a placeholder, so we expect that focuses the editor, and lets the user type.

image

This weird behavior is due to the fact `placheloder` from `draft-js` `Editor` is a `div`
image

So I added a `TouchableWithoutFeedback` that focuses the `Editor` when `onPress`


n0izn0iz (approved):
LGTM

MikaelVallenet (approved):
LGTM, i don't get 100% percent why the placeholder is not a props of the editor/input, but looks fine to me to add a TouchableWithoutFeedback to fix the behavior

PR #1383, opened 2024-11-18
fix: remove dead gno-teritori testing network

n0izn0iz:
this network is down for a long time since it was outdated, keeping it could create confusion


Villaquiranm (approved):

hthieu1110 (approved):
LGTM

Issue #1392, opened 2024-11-18
bug(feed): No loader when react to Gno post


WaDadidou:
This loader should appear
![Image](https://github.com/user-attachments/assets/4f046f2c-b71b-4c5e-a34f-8f8805046a15)

PR #1393, opened 2024-11-18
fix(feed): Add loader when Gno react

WaDadidou:
### Add a loader when react on Gno
Closes https://github.com/TERITORI/teritori-dapp/issues/1392


packages/components/socialFeed/SocialActions/DislikeButton.tsx
```diff
@@ -17,16 +17,16 @@ export const DislikeButton: FC<{
...
     setPost,
   });
 
-  if (isPostMutationLoading)
+  if (isReactLoading)
     return <ActivityIndicator animating color={secondaryColor} size={32} />;
   return (
     <TouchableOpacity
-      onPress={() => handleReaction(DISLIKE_EMOJI)}
+      onPress={() => handleReact(DISLIKE_EMOJI)}
```
MikaelVallenet, line 29: why use `react` insteand of `reaction`.
maybe for new user, at first look it could be confusing with React the library 
Reaction feel more explicit isn't

WaDadidou: It was confusing, I felt like "handleReaction" meant "It fetches reaction". Because the data is named "reactions", so I prefered to use the verb "react", because this hook adds a reaction (the user reacts to a post), and it doesn't fetch data.
But it's not the scope, so I revert this change !
Maybe I'll open a new PR to rename the hook and the function to "useReactToPost" and "reactToPost".

So => https://github.com/TERITORI/teritori-dapp/pull/1393/commits/d9b30c8e9ab5f051c20dc27c71bef2be4ec4cc7f

MikaelVallenet (dismissed):

n0izn0iz (changes requested):
react fails on this PR with
Screenshot 2024-11-20 at 11 09 19

packages/hooks/feed/useSocialReactions.ts
```diff
@@ -61,12 +77,13 @@ export const useSocialReactions = ({
       client,
       msg: {
         icon: emoji,
-        identifier: post.localIdentifier,
+        identifier: "post.localIdentifier",
```
hthieu1110, line 80: do you mean `${post.localIdentifier}` ?

packages/hooks/feed/useSocialReactions.ts
```diff
@@ -75,30 +92,47 @@ export const useSocialReactions = ({
...
+          icon: emoji,
+          count: 1,
+          ownState: true,
+        });
+      }
+      setPost({ ...post, reactions });
+    } catch (err) {
+      console.error(err);
+      if (err instanceof Error) {
+        setToast({
```
hthieu1110, line 124: This is not really in the scope of this PR but I created a related Issue to optimize this code:

https://github.com/TERITORI/teritori-dapp/issues/1401

packages/hooks/feed/useSocialReactions.ts
WaDadidou: Dummy mistake, my bad...
https://github.com/TERITORI/teritori-dapp/pull/1393/commits/0db0ab22f2a177e470025608c3f2948e28016a8e

packages/hooks/feed/useSocialReactions.ts
```diff
@@ -75,6 +81,8 @@ export const useSocialReactions = ({
       func: "ReactPost",
       args: [TERITORI_FEED_ID, post.id.split("-")[1], emoji, "true"],
     };
+
+    setLoading(true);
     const txHash = await adenaDoContract(
```
n0izn0iz, line 86: This can throw, the waitForTransaction also probably

n0izn0iz (changes requested):

packages/hooks/feed/useSocialReactions.ts
WaDadidou: Yes https://github.com/TERITORI/teritori-dapp/pull/1393/commits/269f23fa12b9016eb304b9c97d64bb8a322807ff

n0izn0iz (approved):
LGTM

hthieu1110 (approved):
LGTM

Issue #1367, opened 2024-11-16
bug(feed): dezoomed map by default instead of Paris


WaDadidou:
### We want

![image](https://github.com/user-attachments/assets/2e33b4f5-99f5-4fd6-88d9-d1424c13f587)

### We have

![image](https://github.com/user-attachments/assets/20ce9e0d-e884-4e0e-81a3-df6c4bcf6cf3)

n0izn0iz:
https://github.com/TERITORI/teritori-dapp/pull/1399

PR #1398, opened 2024-11-19
fix(feed): setting zoom level when location not provided

sujal-into:
fix: #1367


packages/components/socialFeed/Map/Map.web.tsx
```diff
@@ -205,8 +206,9 @@ export const Map: FC<MapProps> = ({
         center={
           consultedPostLocation || creatingPostLocation || DEFAULT_MAP_POSITION
         }
-        zoom={12}
+        zoom={!consultedPostLocation && !creatingPostLocation ? 2 : 12}
         attributionControl={false}
+        minZoom={2}
```
WaDadidou, line 211: To have the minimal zoom level and avoid white borders
image

```suggestion
        minZoom={2.15}
```

n0izn0iz: this seem like an arbitrary value, we should be able to compute a proper value from map width and height I think

WaDadidou: You're right
`2.15` is ok with our layout, but the zoom could be more optimal on desktop
The goal is "Compute the smallest `zoom` and `minZoom` value to never see the white background"

packages/components/socialFeed/Map/Map.web.tsx
```diff
@@ -197,6 +197,7 @@ export const Map: FC<MapProps> = ({
           width: "100%",
           height: "100%",
           alignSelf: "center",
+          borderRadius: 20,
```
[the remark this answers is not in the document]
sujal-into: If we provide background color to this view, the background color is seen in the edges. The number is just random, any number above 15 is fine. But I will remove this one as this is not related to this issue.

packages/components/socialFeed/Map/Map.web.tsx
sujal-into: I will implement this in #1399.

packages/components/socialFeed/Map/Map.web.tsx
```diff
@@ -205,8 +206,9 @@ export const Map: FC<MapProps> = ({
         center={
           consultedPostLocation || creatingPostLocation || DEFAULT_MAP_POSITION
         }
-        zoom={12}
+        zoom={!consultedPostLocation && !creatingPostLocation ? 2 : 12}
```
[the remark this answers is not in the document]
WaDadidou: See https://github.com/TERITORI/teritori-dapp/pull/1399 for a dynamic `zoom` value

packages/components/socialFeed/Map/Map.web.tsx
WaDadidou: zoom and bounds are correlated, so ok

WaDadidou (approved):

hthieu1110 (approved):
LGTM

n0izn0iz (changes requested):
please use the maximum responsive size instead of an arbitrary zoom level

sujal-into:
> please use the maximum responsive size instead of an arbitrary zoom level

fixed this in this pr: https://github.com/TERITORI/teritori-dapp/pull/1399

clegirar:
Okay so i'll close this PR and add the changes to the #1399, they are so linked, it will more clear.
And even with the changes, it will still an atomic PR 👍

clegirar:
Okay so i'll close this PR and add the changes to the #1399, they are so linked, it will more clear.
And even with the changes, it will still an atomic PR 👍

Issue #1371, opened 2024-11-16
bug(feed/map): shouldn't be able to zoom out of bounds


n0izn0iz:
we shouldn't be able to zoom out enough to show white stripes

PR #1399, opened 2024-11-19
fix(feed): boundaries and zoom for map feed

sujal-into:
fix: #1371 

Description:
Set the zoom level to 3, so the map gets fit into the container, however, it doesn't show the whole map


n0izn0iz (commented):
surely there must be a way to compute a zoom level that always fit the width of the map no?

it should probably be used in #1398 also

packages/utils/feed/map.ts
```diff
@@ -6,14 +6,22 @@ import musicPostSvg from "@/assets/icons/social-feed-map/music-post.svg";
...
+} from "@/utils/types/feed";
 
 //TODO: Get <svg/> as string from svg files (These svg strings are used in FeedMapList.web.tsx)
 
 export const MAP_LAYER_URL = `https://{s}.tile.jawg.io/jawg-dark/{z}/{x}/{y}{r}.png?access-token=${process.env.EXPO_PUBLIC_LEAFLET_MAP_TOKEN}`;
 
 // Paris baguette
-export const DEFAULT_MAP_POSITION: CustomLatLngExpression = [48.8566, 2.3522];
+export const DEFAULT_MAP_POSITION: CustomLatLngExpression = [20, 0];
+export const MAP_MAX_BOUND: LatLngBoundExpression = [
```
WaDadidou, line 21: Please use `LatLngBoundsLiteral` from `"leaflet`.
And remove that: https://github.com/TERITORI/teritori-dapp/pull/1399/files#diff-573f0234f184f2c9e524405f2c1bb21d17b0f13f1ffcc1b5247d09adec8bca04R77

packages/utils/feed/map.ts
```diff
@@ -6,14 +6,22 @@ import musicPostSvg from "@/assets/icons/social-feed-map/music-post.svg";
...
+  PostCategory,
+} from "@/utils/types/feed";
 
 //TODO: Get <svg/> as string from svg files (These svg strings are used in FeedMapList.web.tsx)
 
 export const MAP_LAYER_URL = `https://{s}.tile.jawg.io/jawg-dark/{z}/{x}/{y}{r}.png?access-token=${process.env.EXPO_PUBLIC_LEAFLET_MAP_TOKEN}`;
 
 // Paris baguette
-export const DEFAULT_MAP_POSITION: CustomLatLngExpression = [48.8566, 2.3522];
+export const DEFAULT_MAP_POSITION: CustomLatLngExpression = [20, 0];
```
WaDadidou, line 20: Please remove remove this change to keep the PR atomic, it's out of scope.
You made this fix in this PR: https://github.com/TERITORI/teritori-dapp/pull/1398/files#diff-a44b6f872b57f327a6fa7ab9ba11100753e475ce41790a0b6b20867e19984addR16

Also, the default position is wrong. But it doesn't concern this PR
image

WaDadidou (changes requested):

packages/utils/feed/map.ts
sujal-into: ok

sujal-into:
> surely there must be a way to compute a zoom level that always fit the width of the map no?
> 
> it should probably be used in #1398 also

I will implement the zoom level logic in the PR of #1398.

n0izn0iz (dismissed):
tested and reviewed

packages/components/socialFeed/Map/Map.web.tsx
```diff
@@ -103,6 +110,43 @@ const MapManager = ({
...
+  const map = useMap();
+
+  useEffect(() => {
+    const lat = 45;
+    const lng = 90;
+
+    const adjustedWorldMapBounds: LatLngBoundsLiteral = [
+      [-lat, -lng],
+      [lat, lng],
+    ];
```
hthieu1110, line 127: This part is independent of the Component  so we should put it outside of the component and we can use UPPER_CASE to define the constants

packages/components/socialFeed/Map/Map.web.tsx
```diff
@@ -103,6 +110,43 @@ const MapManager = ({
...
+
+    const updateMinZoom = () => {
```
hthieu1110, line 129: We should not defined the function inside of useEffect which makes the code less readable.

I suggest sth like
```
useEffect(() => {
    !!map && updateMinZoom(map);
    .......
}, [[map, setMinZoom])
```

hthieu1110 (changes requested):
Could you take a look at my comments pls :) thanks.

packages/components/socialFeed/Map/Map.web.tsx
n0izn0iz: we don't follow this convention, please use normal variable naming as in js, there is no real constants

we can put it outside of the func though

packages/components/socialFeed/Map/Map.web.tsx
n0izn0iz: I disagree, we followed this in berty and it made the code very hard to read IMO

n0izn0iz: you can extract a hook though if you really want

```
const useUpdateMinZoom = (map) => {
    useEffect(...)
}
```

packages/components/socialFeed/Map/Map.web.tsx
sujal-into: fixed here: https://github.com/TERITORI/teritori-dapp/pull/1399/commits/b42b56d5fb01647b17aaa09df59351bb4f9f1532

packages/components/socialFeed/Map/Map.web.tsx
sujal-into: fixed here: https://github.com/TERITORI/teritori-dapp/pull/1399/commits/066f8ccb7aa33572b8b03cb7b1b730e55c4b0ea6

clegirar:
This PR will receive some changes from this PR #1398
I'll explain all changes that i'm going to bring 👍

clegirar:
`MAP_MAX_BOUND` variable is the GPS coordinates (latitude, longitude) of the south-west corner and the north-east corner of a classic planisphere (so they don't come from nowhere)
`zoomSnap` is the value that forces the map's zoom level to always be a multiple of this, so we decreased it to have a better fit with our container
And the last thing I want to mentioned is that the initial leaflet map is square and our container is a rectangle, so this is the main reason we have so much difficulty to fit perfectly with our container, it's because we can't perfectly fit squares into rectangle without have some white strips on the sides

We are obligate to pass zoom initial value to `MapContainer` so it must be arbitrary, we set it to 2.15 (default value that fit with the rectangle in many cases, without white strips on the sides)
I moved updateMinZoom directly in MapManager because it doesn't make a lot of sense to extract a hook that is usable in just one case, and can be write in 6 lines This hook initially did more than 3 lines but the event listener didn't work, so I removed it and I don't understand why we have to handle a resize window behaviour ? I understand we have to handle some responsiveness but resizing windows is not included in that part imo

One thing about the ternary condition in `zoom` props, but the `consultedPostLocation` seems to be not correctly works with `MapContainer`, but I found a props in `setView` for consulted and creation post condition, it's that `setView` take a zoom props that we can directly pass to it, and it works well

Found it little bit confusing to have one value minZoom that is shared for zoom and minZoom value in `MapContainer`, but we don't need to follow the zoom value for the moment so I don't have opinions on that

packages/components/socialFeed/Map/Map.web.tsx
hthieu1110: > we don't follow this convention, please use normal variable naming as in js, there is no real constants
> 
> we can put it outside of the func though

IMHO, if it's a constant and declared outside of function scope, it should be a UPPER_CASE.
if it's just a keyword const and declared in function scope, it's ok to define it with normal case.

n0izn0iz: const in js don't exist
Screenshot 2024-12-04 at 18 02 19

n0izn0iz: this makes refacto harder if the variable stops being a constant according to your definition
also we don't have a way to enforce this rule so this means more review work

n0izn0iz: In the rust language it make more sense because constant are real constant and thus when you see UPPER_CASE, you have some guarantees. Even then it doesn't have much value IMO but at least it can be verified statically.

n0izn0iz: we can vote on this

n0izn0iz:
aussi, en vertu de l'article 49 alinéa 3 de la constitution, j'engage la responsabilité de mon organisation décentralisée sur l'application de la requête de fusionnement numéro 1441

n0izn0iz (approved):
would prefer that it's fully responsive but it's already pretty good

clegirar:
Like even set the boundaries dynamically ?

n0izn0iz:
I mean add back the resize event handler, but it's okay to do it later

n0izn0iz:
qu'ils viennent me chercher

Issue #1370, opened 2024-11-16
bug(feed): MusicAudio and Video don't appear on the Map


WaDadidou:
Examples: 
- Video: https://app.teritori.com/feed/post/testori-d807e0c2-f905-418c-bb75-fe75d18b4a68
- MusicAudio: https://app.teritori.com/feed/post/testori-8abc8bba-7a7d-4932-84da-2dce99fd33c6

But it's ok with VideoNote and Audio

PR #1400, opened 2024-11-20
fix: made keys for social post optional while parsing except title

sujal-into:
fix: #1370

### DIFF
--- a/packages/components/socialFeed/Map/Map.web.tsx
+++ b/packages/components/socialFeed/Map/Map.web.tsx
@@ -134,9 +134,11 @@ export const Map: FC<MapProps> = ({
   const markers: MarkerPopup[] = useMemo(() => {
     if (!posts) return [];
     const results: MarkerPopup[] = [];
-    posts.forEach((post, index) => {
+    posts.forEach((post) => {
       const metadata = zodTryParseJSON(
-        ZodSocialFeedPostMetadata,
+        ZodSocialFeedPostMetadata.partial().merge(
+          ZodSocialFeedPostMetadata.pick({ title: true }),
+        ),
         post.metadata,
       );
       if (!metadata?.location) return;


packages/components/socialFeed/Map/Map.web.tsx
```diff
@@ -134,9 +134,11 @@ export const Map: FC<MapProps> = ({
   const markers: MarkerPopup[] = useMemo(() => {
     if (!posts) return [];
     const results: MarkerPopup[] = [];
-    posts.forEach((post, index) => {
+    posts.forEach((post) => {
       const metadata = zodTryParseJSON(
-        ZodSocialFeedPostMetadata,
+        ZodSocialFeedPostMetadata.partial().merge(
+          ZodSocialFeedPostMetadata.pick({ title: true }),
```
WaDadidou, line 140: You don't need to modify the Zod type to try parse.

The issue is due to the fact we have 4 differents types for the fetched posts.
Here we only try to parse `ZodSocialFeedPostMetadata`. So, the tracks and videos posts validation fails.

The `result` here has no `success`: https://github.com/TERITORI/teritori-dapp/blob/bugfix/video-audio-map/packages/utils/sanitize.ts#L22
If you `console.log` the `result`, you will see what's wrong to validate `ZodSocialFeedPostMetadata`

To avoid that, because the map gathers all posts categories, we must handle these Zod types. Here is my suggestion:

```
const videoPostMetadata = zodTryParseJSON(
  ZodSocialFeedVideoMetadata,
  post.metadata,
);
const trackMetadata = zodTryParseJSON(
  ZodSocialFeedTrackMetadata,
  post.metadata,
);
const articleMetadata = zodTryParseJSON(
  ZodSocialFeedArticleMetadata,
  post.metadata,
);
const postMetadata = zodTryParseJSON(
  ZodSocialFeedPostMetadata,
  post.metadata,
);
const metadataToUse = videoPostMetadata || trackMetadata || articleMetadata || postMetadata;
```

We need a refacto about it. Either unify all types in one, either add an utilitary function to handle (try parse) these differents zod types

WaDadidou (changes requested):

packages/components/socialFeed/Map/Map.web.tsx
n0izn0iz: actually no there is the `zodSocialFeedCommonMetadata` when you need to only parse the title

sujal-into: I have made a fix for this discussion in this commit: https://github.com/TERITORI/teritori-dapp/pull/1400/commits/3a16113c647d667b39245f99481859b42872c99b

sujal-into: > actually no there is the `zodSocialFeedCommonMetadata` schema when you need to only parse the title

In this case, there are other things that i need to parse also based on the category type, so I have created a util function to handle those cases.

n0izn0iz: you can switch on the post type to do the proper parsing in each case, the whole point of using zod is to have type safety, using a `z.ZodTypeAny` entirely defeat this purpose

n0izn0iz (changes requested):
no `z.ZodTypeAny` please

packages/utils/types/feed.ts
```diff
@@ -135,3 +136,19 @@ export type ReplyToType = {
...
+
+export const parseSocialFeedMetadata = (
+  category: PostCategory,
+  metadata: string,
+) => {
+  switch (category) {
+    case PostCategory.Video:
+      return zodTryParseJSON(ZodSocialFeedVideoMetadata, metadata);
+    case PostCategory.Article:
+      return zodTryParseJSON(ZodSocialFeedPostMetadata, metadata);
```
WaDadidou, line 148:
```suggestion
      return zodTryParseJSON(ZodSocialFeedArticleMetadata, metadata);
```

WaDadidou (changes requested):

packages/utils/types/feed.ts
```diff
@@ -135,3 +136,19 @@ export type ReplyToType = {
...
+    case PostCategory.MusicAudio:
+      return zodTryParseJSON(ZodSocialFeedTrackMetadata, metadata);
+    default:
+      return zodTryParseJSON(ZodSocialFeedPostMetadata, metadata);
+  }
```
WaDadidou, line 153:
```suggestion
  switch (category) {
    case PostCategory.Video:
      return zodTryParseJSON(ZodSocialFeedPostMetadata, metadata) || zodTryParseJSON(ZodSocialFeedVideoMetadata, metadata);
    case PostCategory.Article:
      return zodTryParseJSON(ZodSocialFeedPostMetadata, metadata) || zodTryParseJSON(ZodSocialFeedArticleMetadata, metadata);
    case PostCategory.MusicAudio:
      return zodTryParseJSON(ZodSocialFeedTrackMetadata, metadata);
    default:
      return zodTryParseJSON(ZodSocialFeedPostMetadata, metadata);
  }
```
- Use ZodSocialFeedArticleMetadata for Articles
- Hanlde the "old" posts. Example: https://github.com/TERITORI/teritori-dapp/blob/bugfix/video-audio-map/packages/components/socialFeed/Map/MapPosts/ArticleMapPost.tsx#L24-L26

We'll make another PR to replace all `zodTryParseJSON` usages by  `parseSocialFeedMetadata` (About posts)

WaDadidou (changes requested):

packages/utils/types/feed.ts
n0izn0iz: > Hanlde the "old" posts

actually we will reset the feed after redesigning the contracts so I don't think it's important to support old stuff right now, even all this refacto is not necessary since we'll have well defined types from contracts

packages/utils/types/feed.ts
n0izn0iz, line 148:
```suggestion
      return zodTryParseJSON(ZodSocialFeedArticleMetadata, metadata);
```

packages/utils/types/feed.ts
```diff
@@ -135,3 +136,19 @@ export type ReplyToType = {
...
+) => {
+  switch (category) {
+    case PostCategory.Video:
+      return zodTryParseJSON(ZodSocialFeedVideoMetadata, metadata);
+    case PostCategory.Article:
+      return zodTryParseJSON(ZodSocialFeedPostMetadata, metadata);
+    case PostCategory.MusicAudio:
+      return zodTryParseJSON(ZodSocialFeedTrackMetadata, metadata);
+    default:
+      return zodTryParseJSON(ZodSocialFeedPostMetadata, metadata);
```
n0izn0iz, line 152:
```suggestion
      return zodTryParseJSON(ZodSocialFeedCommonMetadata, metadata);
```

n0izn0iz (changes requested):
if you need other cases, add them, there should be one zod schema for one post category and the fallback is common metadata

packages/utils/types/feed.ts
sujal-into: I made changes here: https://github.com/TERITORI/teritori-dapp/pull/1400/commits/8a0ebc639cf8ad35aee3d70b4fa098f3a0bfd133

WaDadidou: @n0izn0iz Yes, some issues are "stand by" for now

packages/utils/types/feed.ts
```diff
@@ -135,3 +136,27 @@ export type ReplyToType = {
...
+export const parseSocialFeedMetadata = (
+  category: PostCategory,
+  metadata: string,
+) => {
+  switch (category) {
+    case PostCategory.Video:
+      return (
+        zodTryParseJSON(ZodSocialFeedPostMetadata, metadata) ||
+        zodTryParseJSON(ZodSocialFeedVideoMetadata, metadata)
+      );
```
n0izn0iz, line 149:
```suggestion
      return (
        zodTryParseJSON(ZodSocialFeedVideoMetadata, metadata)
      );
```

packages/utils/types/feed.ts
```diff
@@ -135,3 +136,27 @@ export type ReplyToType = {
...
+    case PostCategory.Article:
+      return (
+        zodTryParseJSON(ZodSocialFeedPostMetadata, metadata) ||
+        zodTryParseJSON(ZodSocialFeedArticleMetadata, metadata)
+      );
```
n0izn0iz, line 154:
```suggestion
      return (
        zodTryParseJSON(ZodSocialFeedArticleMetadata, metadata)
      );
```

n0izn0iz (changes requested):

packages/utils/types/feed.ts
sujal-into: Resolved it here: https://github.com/TERITORI/teritori-dapp/pull/1400/commits/0b9046c2de0b394297ca278952dff0b02b399731

packages/utils/types/feed.ts
sujal-into: Resolved  it  here: https://github.com/TERITORI/teritori-dapp/pull/1400/commits/0b9046c2de0b394297ca278952dff0b02b399731

WaDadidou:
So, we pause that, we need refacto

Issue #1379, opened 2024-11-18
chore(feed): Slider floods browser console with StyleSheet warnings


WaDadidou:
Caused by `import Slider from "@react-native-community/slider";`

![image](https://github.com/user-attachments/assets/8897c62a-7c8a-433c-a8bd-a14a1a40e659)

PR #1406, opened 2024-11-21
fix(feed:) Update slider package to prevent flooding warnings

WaDadidou:
Closes https://github.com/TERITORI/teritori-dapp/issues/1379

### Why this PR ?
image

### The react-native-slider fix:
https://github.com/callstack/react-native-slider/releases/tag/v4.5.3

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
     "@react-native-anywhere/polyfill-base64": "^0.0.1-alpha.0",
     "@react-native-async-storage/async-storage": "1.21.0",
     "@react-native-clipboard/clipboard": "^1.13.2",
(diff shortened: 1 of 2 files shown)


n0izn0iz:
need to upgrade expo, this is not a light task, I think we should not tackle that right now

WaDadidou:
> need to upgrade expo, this is not a light task, I think we should not tackle that right now

Why expo ?

n0izn0iz:
https://github.com/TERITORI/teritori-dapp/actions/runs/11959178214/job/33340393703?pr=1406#step:5:35

hthieu1110 (dismissed):
LGTM

PR #1412, opened 2024-11-22
fix: dao proposals query key

n0izn0iz:
without this, if we query the proposals for a teritori dao and the gno query resolves first with nothing, the teritori query will return an empty list

### DIFF
--- a/packages/hooks/dao/useDAOProposals.ts
+++ b/packages/hooks/dao/useDAOProposals.ts
@@ -51,7 +51,7 @@ export const useDAOProposals = (daoId: string | undefined) => {
   const { daoProposals: cosmWasmDAOProposals, ...cosmWasmOther } =
     useCosmWasmDAOProposals(daoId);
   const { data: gnoDAOProposals, ...gnoOther } = useQuery(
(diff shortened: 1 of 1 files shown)


hthieu1110 (dismissed):
as the name of the hook is `useDAOProposals` which is sth generic, not dedicated to Gno so I suggest:
- either keep the hook generic name and make all the query keys (string "gno" => networkId) more generic
- or change the name to useGnoDAOProposal

but not mix both generic name + hardcoded network => quite confusing

hthieu1110 (commented):
as the name of the hook is useDAOProposals which is sth generic, not dedicated to Gno so I suggest:

either keep the hook generic name and make all the query keys (string "gno" => networkId) more generic
or change the name to useGnoDAOProposal
but not mix both generic name + hardcoded network => quite confusing

n0izn0iz:
I think you misread the diff, the hook is indeed generic, it's just that the gno version is not extracted where the cosmwasm one is

It was a quickfix but I can refacto to improve clarity

n0izn0iz:
merging because this bug is very anoying

Issue #1409, opened 2024-11-22
fix(top-menu): Replace the banner by the news


WaDadidou:
Bellow, Here is a screenshot of the Home screen with the TopMenu open.
In the TopMenu, at "Highlighted News", the Home banner is shown instead of the news, that's wrong.
The news to use are the same data than on the Home page: https://github.com/TERITORI/teritori-dapp/blob/chore-thumbnail-media/packages/components/carousels/NewsCarouselSection.tsx#L22

You have to :

- [ ] Fetch the news instead of the banners here: https://github.com/TERITORI/teritori-dapp/blob/chore-thumbnail-media/packages/components/TopMenu/TopMenuHighlightedNews.tsx#L14
- [ ] Make a small version of this carousel: https://github.com/TERITORI/teritori-dapp/blob/chore-thumbnail-media/packages/components/hub/HubLanding.tsx#L55
- [ ] Place it here: https://github.com/TERITORI/teritori-dapp/blob/chore-thumbnail-media/packages/components/TopMenu/TopMenuHighlightedNews.tsx#L19-L34
- [ ] Plug your carousel with the fetched news

Take attention to the existing components before making new ones, and ask information if you need :)

![Image](https://github.com/user-attachments/assets/eda19229-4146-4bba-a364-98dbb01807d2)
- Red: Current wrong state
- Green: What it should be

Dhir0808:
Hello WaDadidou, I would like to take this issue.

WaDadidou:
Hello @apoorvapendse It could be tricky, but if you are not affraid by our front, let's try :)
When it's done, make a PR and mention this issue

PR #1417, opened 2024-11-25
fix: Replace home banner with news carousel.

apoorvapendse:
This pull request fixes the issue of displaying home banner instead of a news carousel on the top menu.
Fixes: #1409 

**Video Demo:**
https://github.com/user-attachments/assets/43ac97a7-cae1-434d-8fb0-e153bdd0e060

### DIFF
--- a/packages/components/TopMenu/TopMenuHighlightedNews.tsx
+++ b/packages/components/TopMenu/TopMenuHighlightedNews.tsx
@@ -1,37 +1,80 @@
-import React from "react";
-import { Image } from "react-native";
+import React, { useEffect, useRef } from "react";
+import { TouchableOpacity, View } from "react-native";
+import Carousel, { ICarouselInstance } from "react-native-reanimated-carousel";
 
 import { TopMenuSection } from "./TopMenuSection";
-import { useBanners } from "../../hooks/marketing/useBanners";
+import chevronLeftSVG from "../../../assets/icons/chevron-left.svg";
+import chevronRightSVG from "../../../assets/icons/chevron-right.svg";
 import { useSelectedNetworkId } from "../../hooks/useSelectedNetwork";
-import { web3ToWeb2URI } from "../../utils/ipfs";
-import FlexCol from "../FlexCol";
-import { Link } from "../Link";
-import { PrimaryBox } from "../boxes/PrimaryBox";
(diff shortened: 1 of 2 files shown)


WaDadidou:
As we don't have design for that, we expect you to purpose a design for the TopMenu version of the News carousel.
I know it's not really clear, you need more information to handle this issue.

Here is a guideline: 
- Blue: The big one is in an horizontal rectange, you must keep a horizontal rectangle shape (approximative) to the TopMenu
- Green: Just show the `title` and `subtitle`on top left, the buttons CTAs on bottom left, and the image on right
- Red: Don't show the `text`
- Obvisouly, reduce the image height/width, the title and subtitle height/width. You reduced the title size, but not the subtitle size. The title must be bigger than the subtitle, but both must be smaller (Not too smaller, refer to the other font sizes used in TopMenu)
- I recommend you to make a `TopMeuNewsBox` and use it in `TopMenuHighligjtedNews`.
$image

About your design:
- It takes too much space, 
- The buttons are too big, we already have buttons in TopMenu, use the same size
- There are vertical spaces above and bellow
- The left/right buttons must be at the right of the title "Highlighted News", on the same line, like the big one on the home page

image

WaDadidou:
Sorry for the lack of information, we are kinda busy, but I want to test colaborating with external people.
I modified the tasks on the issue: https://github.com/TERITORI/teritori-dapp/issues/1409

WaDadidou:
Hello @apoorvapendse ,
I understand the confusion ^^"
The height must be fix. 
Exemple: If a News has no subtitle, there will be an empty space. If the News has a big title or subtitle, it will be cropped (See `numberOfLines` props of `BrandText`).

Regarding your last video, you can:
 - Reduce the subtitle font size
 - Reduce the image size
 
 Here is the layout that I think is good:
image

🟦 Blue: "Highlighted News" + Left/Right buttons horizontally
🔴 Red: Title + Subtitle (2 lines maximum for title and 2 lines maximum for subtitle)
🟨 Yellow: Image
🟢 Green: Buttons CTAs horizontally

image

 Maybe around 340px as max height for the whole the could be good.

___
Thank you for your exploration, please try to integrate this layout and push your code :)
I take care about this issue, and I train myself in communication

Issue #1380, opened 2024-11-18
bug(wallets/adena): switching network in dapp does not properly asks to switch network in adena


n0izn0iz:
the network switching window is opened and then closed instantly

n0izn0iz:
@hthieu1110 please properly use [keywords](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests#linking-a-pull-request-to-an-issue) to autoclose issues when merging PRs

PR #1424, opened 2024-11-26
fix: adena network sync

hthieu1110:
Fix issue: https://github.com/TERITORI/teritori-dapp/issues/1380


n0izn0iz (changes requested):
Looks good, except there is a weird behavior when you connect for the first time (when the app is not authorized yet) and you are on the wrong gno network (for example, adena is on test5 and the app is on portal loop)

Screenshot 2024-11-26 at 14 06 49

n0izn0iz (changes requested):
still have the same behavior, please read the previous comment carefully to reproduce

WaDadidou:
### I have this: 
image

### When I click Connect, I have an infinite loading:
image

### When I close the wallet window, I got this error : 
`Uncaught (in promise) TypeError: Cannot read properties of null (reading 'chainId')`

Same on deploy-preview-1424--gno-dapp et deploy-preview-1424--teritori-dapp

hthieu1110:
> Same on deploy-preview-1424--gno-dapp et deploy-preview-1424--teritori-dapp:
> 
> ### I have this:
> image
> 
> ### When I click Connect, I have an infinite loading:
> image
> 
> ### When I close the wallet window, I got this error :
> `Uncaught (in promise) TypeError: Cannot read properties of null (reading 'chainId')`

I'm not able to reproduce this. Step I've tried:
- Open the dapp with unauthorized wallet
- Select Gno Test 5 while in Adena it's Portal loop
- Connect with Adena
=> it shows only the error that Norman mentioned above

hthieu1110:
> still have the same behavior, please read the previous comment carefully to reproduce
> 
> you need to remove the authorization in this list to reproduce
> 
> Screenshot 2024-11-26 at 18 15 46

Okie, I've push the fix and cover some other edge-cases also.

hthieu1110:
> @hthieu1110 I got this error when the Adena Switch modal is open, BUT I click on "Connect Wallet" then "Adena Wallet"
> 
> image
> 
> I think we should alway show the loader in ConnectWalletButton when a wallet window in open, to prevent the user from breaking the UI if he doesn't interact with the wallet extension.
> 
> image
> 
> 
> 
> Maybe out of scope
> 
> 
> 
> 

IMHO, it's a discussion in another scope, for now, when the switch modal is open, you ask to connect to another Adena network so it's normal that it fetches the connected network from your wallet and thus close the modal. That behaviour is not really related to this Pr.

n0izn0iz:
actually @WaDadidou is right, this is why we were calling `getAccount` in the button handler

but fixing this issue is more important than the problem of introducing this regression so we can merge as-is and open another issue IMO

n0izn0iz (changes requested):
Following the same flow as before (first connection, wrong network):
if you cancel the "switch network" request in adena, the "Connect wallet" button then does nothing

n0izn0iz:
I recommend you try testing in "adversarial mode", thinking about way to break the flow before asking for new reviews

hthieu1110:
> I recommend you try testing in "adversarial mode", thinking about way to break the flow before asking for new reviews

Of course testing all the flow possible on my side is what I have done firstly before asking the review. I've tested all the cases that I think about and there are edge cases that I dont even know

WaDadidou:
Empty app's local data
Wallet: Portal Loop
https://deploy-preview-1424--gno-dapp.netlify.app/orgs

image

==> I got empty netowrks list, even after Click connect, then connect adena
image
image

WaDadidou:
- Empty app's local data
- Wallet: Portal Loop
- Cancel the Switch from Adena modal
https://deploy-preview-1424--gno-dapp.netlify.app/feed

image

==> I got the expected error
image

n0izn0iz (changes requested):
Now the same flow I described just doesn't try to change network in adena
This can result in the following for pages where there is only one valid gno network for the page like a [post details page](https://deploy-preview-1424--gno-dapp.netlify.app/user/gnotest5-g199v9fxazlp2n7ls7n4e04x0q4rcg3ggehtjdsh)

Screenshot 2024-11-29 at 14 45 14

WaDadidou:
I'm a little bit lost with the flows to tests, I'm testing following the changes and report I I find something. And report the expected cases to make them clear.
But it looks good, I didn't get issue.

I saw this expected case:
image

WaDadidou:
### I had some weird behaviors:
- "The account is already connected to this website" if I:
   - Lock Adena wallet
   - Refresh the page (And got "Adena is locked")
   - Click Connect Wallet on Teritori, Click on Adena Wallet on the modal on Teritori
- "Switching to the network has been rejected by the user" if I:
   - Click Connect Wallet from Teritori instead of Switch Network from Adena modal
   
### We must lock the Connect Wallet button' by disabling the button and/or showing a loader or a button text "Look at ur Adena extension plz". Weird behaviors could be prevent


https://github.com/user-attachments/assets/444e8986-0d1e-496c-bd89-a580a60e8bb9

### Random tests, not really important, just got a weid behavior at the end, but I due to all side effects I guess

https://github.com/user-attachments/assets/887dc853-398b-4020-b380-cd974154aa05

n0izn0iz:
I get a `Adena is locked` error toast without any interaction on my side when I navigate to the site

WaDadidou:
> I get a `Adena is locked` error toast without any interaction on my side when I navigate to the site

You should have this state on your Adena wallet, is it the case ?

image

packages/context/WalletsProvider/adena/index.ts
```diff
@@ -20,90 +25,86 @@ export const useAdena: () => UseAdenaResult = () => {
...
-    return () => window.removeEventListener("load", handleLoad);
-  }, []);
+  const { addAdenaNetwork, switchAdenaNetwork } = useAdenaUtils();
 
-  useEffect(() => {
-    const effect = async () => {
-      if (!hasAdena || !isAdenaConnected) {
+  const fetchAccount = useCallback(
+    async (
+      dispatch: any,
```
n0izn0iz, line 39: don't use any for this

n0izn0iz (changes requested):

packages/context/WalletsProvider/adena/index.ts
hthieu1110: Fixed

n0izn0iz:
aussi, en vertu de l'article 49 alinéa 3 de la constitution, j'engage la responsabilité de mon organisation décentralisée sur l'application de la requête de fusionnement numéro 1424

PR #1425, opened 2024-11-28
fix: removed unused effect

addegbenga:
Remove unused useEffect from FeedPostDefaultView screen
Fixes #1350

I just opened a new PR with the exact same changes to get CI Right.
[PR](https://github.com/TERITORI/teritori-dapp/pull/1427)

PR #1427, opened 2024-11-28
fix: removed unused effect

### LINKED ISSUE #1350: chore(feed): Remove this useless useEffect used for Video render
It's here: https://github.com/TERITORI/teritori-dapp/blob/main/packages/screens/FeedPostView/components/FeedPostDefaultView.tsx#L57-L62

Why removing that ?

`FeedPostDefaultView` is only called in `FeedPostView`, but this `useEffect` redirects to `FeedPostView`.
Also, the `PostCategory.Video` case is already handled in `FeedPostView`: https://github.com/TERITORI/teritori-dapp/blob/main/packages/screens/FeedPostView/FeedPostView.tsx#L81-L82.
So, there is no reason to redirect to `FeedPostView` for the `PostCategory.Video` case

addegbenga:
Remove unused useEffect from FeedPostDefaultView screen
Fixes #1350

### DIFF
--- a/packages/screens/FeedPostView/components/FeedPostDefaultView.tsx
+++ b/packages/screens/FeedPostView/components/FeedPostDefaultView.tsx
@@ -54,13 +54,6 @@ export const FeedPostDefaultView: FC<{
   const isMobile = useIsMobile();
   const [parentOffsetValue, setParentOffsetValue] = useState(0);
 
-  useEffect(() => {
-    if (post.category === PostCategory.Video)
-      navigation.replace("FeedPostView", {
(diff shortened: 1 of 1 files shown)


clegirar:
Old PR is this one https://github.com/TERITORI/teritori-dapp/pull/1425
So we LGTM with @WaDadidou 👍

clegirar (approved):
LGTM 👍

MikaelVallenet (approved):

hthieu1110 (approved):
LGTM

PR #1429, opened 2024-11-29
feat(orgs): split creation flows + add new kind of roles based organization

MikaelVallenet:
## Integration of Roles Manager into Organizations Flow

The roles manager allow to manage roles & permissions on a user base.
This PR integrates it into our DAO gno contracts

Exhaustive list of additions:
- Add a new Module called ``RolesModule`` within DAO Core with its own interface see ``gno/p/dao_interfaces``
- add a new gno pure package ``dao_roles_group`` that implements the interface ``IRolesModule``
- Add these functions: ``IsRoleExist``, ``CountRoles``, ``GetUserRoles(user)`` into ``role manager`` pure pkg
- Refactor Organization flow to split the logic (except the first step that is shared)
- Remove token based logic (was not working for now anyway)
- Reset form data on structure type change & on tx cancel
- Reset unlocked step on structure type change
- Add a roles setting step for roles based struct with unlimited list of roles with color
- Add an input to set roles to members in roles based dao
- Adapt the generated gno contract to work roles based dao
- Refactor project structure for e2e gno tests (out of scope, but since it's a small adjustement i don't think i should open an atomir PR just for this, waste of time imo)
- Add of register the dao profile (displayName, description, ImageURI) => i will move this to an atomic PR
- move ``getDuration`` && ``getPercent`` into ``utils/gnodao/helpers.ts``
- Remove and replace all stylesheet.Create in ``organizations/**``

> Membership based and Roles dao based share the same contracts, just the membership are roles dao based organization without any roles


n0izn0iz:
> I've decided to create an “organizations-role-manager” branch, which will receive atomic PRs from now on to gradually reach a usable stage, at which point it will merge with main.

We had a lot of problems with long-running branches and decided to avoid them so I would advise against that and instead to try make atomic PRs against `main`

n0izn0iz:
maybe we could start by a refactor of the org creation flow and introduce a new flow that doesn't affect the existing.
- add a new "dao structure" in the ui
- hide it in non-dev mode
this way you can iterate on it on main, no?

packages/screens/Organizations/components/CreateDAOSection.tsx
```diff
@@ -49,6 +51,7 @@ export const CreateDAOSection: React.FC<CreateDAOSectionProps> = ({
   const selectedRadioStructure = watch("structure");
   const uri = watch("imageUrl");
   const name = watch("associatedHandle");
+  const isDev = process.env.NODE_ENV === "development";
```
n0izn0iz, line 54: See https://github.com/TERITORI/teritori-dapp/blob/main/CONTRIBUTING.md#developer-mode
```suggestion
  const [isDev] = useDeveloperMode();
```

n0izn0iz (changes requested):

gno/r/dao_realm/dao_realm.gno
```diff
@@ -73,7 +87,12 @@ func init() {
 		},
 	}
 
-	daoCore = dao_core.NewDAOCore(votingModuleFactory, proposalModulesFactories, messageHandlersFactories)
+	daoCore = dao_core.NewDAOCore(votingModuleFactory, rolesModuleFactory, proposalModulesFactories, messageHandlersFactories)
+
+	// Register the DAO profile
+	profile.SetStringField(profile.DisplayName, "DAO Realm")
+	profile.SetStringField(profile.Bio, "Default testing DAO")
+	profile.SetStringField(profile.Avatar, "")
```
MikaelVallenet, line 95: Is not handled in another PR already ? could split it up in another PR

networks.json
```diff
@@ -4581,6 +4583,64 @@
     "profilePkgPath": "gno.land/r/demo/profile",
     "txIndexerURL": "https://indexer.portal-loop.gno.testnet.teritori.com"
   },
+  {
+    "id": "gno-teritori",
```
MikaelVallenet, line 4587: remove this by regenerating networks.json after merged main branch

packages/components/user/UserCard.tsx
```diff
@@ -114,48 +115,48 @@ export const UserCard: React.FC<{
...
-              </View>
-            );
-          })}
-        </ScrollView>
+        {roles && (
+          <View>
+            <BrandText
+              style={[fontSemibold12, { lineHeight: 14, marginBottom: 8 }]}
+            >
+              Roles
```
MikaelVallenet, line 123: if roles is empty don't display the text

packages/screens/Organizations/components/MembershipOrg/MembershipMemberSettingsSection.tsx
```diff
@@ -21,13 +21,13 @@ import {
   ORGANIZATION_DEPLOYER_STEPS,
 } from "@/utils/types/organizations";
 
-interface MemberSettingsSectionProps {
+interface MembershipMemberSettingsSectionProps {
```
MikaelVallenet, line 24: i don't like the naming but i can't think of something that answer this:
- should describe the kind of org, here a membership based org
- should describe the section here : Member Settings

packages/screens/Organizations/components/ReviewInformationSection.tsx
```diff
@@ -44,6 +46,7 @@ export const ReviewInformationSection: React.FC<
   votingSettingData,
   tokenSettingData,
```
MikaelVallenet, line 47: delete this file

packages/utils/types/organizations.ts
```diff
@@ -22,22 +27,41 @@ export type ConfigureVotingFormType = {
...
+
+export type RolesSettingFormType = {
+  roles: { name: string; color: string }[];
 };
 
-export type LaunchingProcessStepType = {
-  title: string;
-  completeText: string;
-  isComplete?: boolean;
+// SPLIT IT INTO ROLES & NORMAL MEMBERSHIPS ORGS
```
MikaelVallenet, line 50: so do it now

gno/r/dao_realm/dao_realm.gno
MikaelVallenet: my bad i put it two times (should come from a failed merged)

gno/p/dao_interfaces/modules.gno
```diff
@@ -37,3 +37,17 @@ type IProposalModule interface {
 }
 
 type ProposalModuleFactory func(core IDAOCore) IProposalModule
+
+type IRolesModule interface {
```
hthieu1110, line 41: IMHO, it could be nice and more practice if we handle AddressOrName instead of only Address

n0izn0iz: I disagree, if we do we have to add a way to resolve users by passing a resolver in the role module, increasing gas cost, dependencies and complexity
then if at some point usernames can change, we get a lot of headache
IMO it's better to always operate on addresses where possible

clegirar:
Screenshot 2024-12-04 at 21 57 53
A little confusing fontSize here

Screenshot 2024-12-04 at 22 19 10
I have 2 things to say here, thrash icons are too big imo, and maybe you can fit the thrash icon with the end of the container ? Like a full width
In term of components behavior, on the configuring voting tab, all `<TextInputCustom<ConfigureVotingFormType>` are not clickable and just scrollable (out of scope i think btw), i think the best is to have both possibilities

Screenshot 2024-12-04 at 22 21 19
In the recap at the end of the form, just before the creation maybe we can have the colors for each role ?

Screenshot 2024-12-04 at 22 36 49
At the end of the creation, after the validated animation, the middle circle stay here like that ? Maybe this behaviour was already here before (I'll check, but found that weird, and in the same time we have a call-to-action component, so idk maybe i'm tickling)

Screenshot 2024-12-04 at 22 40 02
Maybe create a `role` tab for a role based DAO ?

MikaelVallenet:
> Screenshot 2024-12-04 at 21 57 53 A little confusing fontSize here
> 
> Screenshot 2024-12-04 at 22 19 10 I have 2 things to say here, thrash icons are too big imo, and maybe you can fit the thrash icon with the end of the container ? Like a full width In term of components behavior, on the configuring voting tab, all `<TextInputCustom<ConfigureVotingFormType>` are not clickable and just scrollable (out of scope i think btw), i think the best is to have both possibilities
> 
> Screenshot 2024-12-04 at 22 21 19 In the recap at the end of the form, just before the creation maybe we can have the colors for each role ?
> 
> Screenshot 2024-12-04 at 22 36 49 At the end of the creation, after the validated animation, the middle circle stay here like that ? Maybe this behaviour was already here before (I'll check, but found that weird, and in the same time we have a call-to-action component, so idk maybe i'm tickling)
> 
> Screenshot 2024-12-04 at 22 40 02 Maybe create a `role` tab for a role based DAO ?

Thanks you a lot for the review :pray: 
I will look for the font size, for the rest i would prefer to focus on continue building instead of invest time on it.
For the role tab it's plan in the next PRs
I propose i open issues about what you said after this PR merge so i can do atomic PR later :heart:

PR #1436, opened 2024-12-03
feat: improve app switch logic and add logo customization

n0izn0iz:
- Improved the app switch to not edit `package.json`
- Added a way to customize the top logo and added gnotribe tmp top logo
- Added gnotribe icon


hthieu1110:
Should we add app-switch.config.example to the repo ? if not then someone later has to ask how it looks like the content of app-switch :-s

n0izn0iz:
why? it's committed as you can see in the diff

n0izn0iz:
actually this creates a diff unless you run specific update-index commands locally :/
also the android build seem to do funny business with the app config so the switch doesn't work

MikaelVallenet (approved):
I approve because the working flow looks good to me,
Some idea, what about renaming ``gno-dapp`` -> ``gnotribe-dapp``, it's more explicit since the product is name gnotribe

could also be nice in a script when an app is not found to list the existing app (just list name of folders in ``apps`` to get this informations

clegirar (approved):
Tested it and it works well 👍 
Cool to have an alternative without change `package.json`

And the error message with available apps is good

MikaelVallenet (approved):

PR #1440, opened 2024-12-04
fix: fix types for ts client for social-feed

hthieu1110:
NOTE:
This fix the types for ts client for teritori-social-feed, due to updated contract is missing so this needs some manual tweak to make it works (and compatible with lint).


n0izn0iz:
we're not sure of the actual types and we will refacto the contracts so I'm closing this

PR #1446, opened 2024-12-05
fix: reduce fontSize and fontWeight of all title screens

clegirar:
- Add new component `ScreenTitle`
- Use `ScreenTitle` for every screen title (in `ScreenContainer`, in the `headerChildren` props)
- Reduced default fontSize and fontWeight for all title screens

### DIFF
--- a/packages/components/ScreenContainer/ScreenTitle.tsx
+++ b/packages/components/ScreenContainer/ScreenTitle.tsx
@@ -1,11 +1,15 @@
-import { TextProps } from "react-native";
+import { ReactNode } from "react";
 
 import { BrandText } from "../BrandText";
 
 import { fontRegular15 } from "@/utils/style/fonts";
 
-export const ScreenTitle: React.FC<TextProps> = (props) => {
+export const ScreenTitle: React.FC<{
+  children?: ReactNode;
+}> = ({ children }) => {
   return (
-    <BrandText style={[fontRegular15, props.style]}>{props.children}</BrandText>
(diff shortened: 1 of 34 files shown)


packages/components/ScreenContainer/ScreenTitle.tsx
```diff
@@ -0,0 +1,11 @@
+import { TextProps } from "react-native";
+
+import { BrandText } from "../BrandText";
+
+import { fontRegular15 } from "@/utils/style/fonts";
+
+export const ScreenTitle: React.FC<TextProps> = (props) => {
```
n0izn0iz, line 7:
```suggestion
export const ScreenTitle: React.FC<{style: StyleProp<TextStyle>; children: React.Node}> = (props) => {
```

n0izn0iz (changes requested):

n0izn0iz:
aussi, en vertu de l'article 49 alinéa 3 de la constitution, j'engage la responsabilité de mon organisation décentralisée sur l'application de la requête de fusionnement numéro 1446

packages/components/ScreenContainer/ScreenTitle.tsx
```diff
@@ -0,0 +1,15 @@
+import { ReactNode } from "react";
+
+import { BrandText } from "../BrandText";
+
+import { fontRegular15 } from "@/utils/style/fonts";
+
+export const ScreenTitle: React.FC<{
```
hthieu1110, line 7: According to all the changes I see in this PR, the children is always text and in no case children can be undefined. So is that OK to change children to mandatory props and children become string instead of ReactNode ?

PR #1447, opened 2024-12-05
feat(feed): take maximum width

clegirar:
Initially i started to work on a fullWidth screen, but i encounter some issues and reflexions that i want to share with you.

I found `ScreenContainer` component A LOT confusing, can we just choose one layout for container and have it in all of screens ?
 Else we end up with this kind of "generic" component, that have a lot of props, a lot of conditions and a lot of behaviours.. complicated to maintain, complicated to change etc.
 
 We have some props that are not very true (like `responsive` one, the only thing it did is call a function that is just arbitrary and not responsive https://github.com/TERITORI/teritori-dapp/blob/main/packages/utils/style/layout.ts#L18.
 
`hideSidebar` props is used only in `ToryPunks` screen, it's specific, can we show `Sidebar` in any case and just fit with all screens behaviour ? 
I even want to have the same header children for all of screens, and just pass a title to the ScreenContainer , wdyt ?
Have we to maintain `Mini` ? So the mobile version ? If yes why we pass through `ScreenContainer` to finally call `ScreenContainerMobile`, and not directly `ScreenContainerMobile` in all mobiles screens ?

And a last thing, can I have more informations about `forceNetworkId`, `forceNetworkKind` and `forceNetworkFeatures` ? Are they still necessary ? What are their role ?

I propose a new component that is similar in all screens, with similar margin, that scrolls if content are bigger than the height of the screen (not implemented yet), that don't have `backPress` option, that don't have all of this "generic" props, that makes this component confusing.

Right now his name is `NewScreenContainer`, i'll try to reduce containers levels inside this component like the FIXME comment said 👍 
It's still WIP obviously so the code is not clean etc. i just want to start the discussion.

Tell me what you think about that 👍
Thanks to you by advance (and just to remind, i don't blame anybody or anything, just questions and comments ❤️)

Little preview
Screenshot 2024-12-05 at 23 30 01
(We can see in this screen that some components are not compatible with a full width screen, like audio posts)


n0izn0iz:
Screenshot 2024-12-06 at 07 29 24

clegirar:
> IMO we should simplify the existing one instead of creating a new one and increasing complexity

If we create a new component and we replace it everywhere what is the problem for you ?
In fact when you check my component it's a pure copy paste of the actual one, without all of props i found "generic" or props that need special conditions, that make the component complex.
So i can just simplify the actual one.

---
> There is more logic in useMaxResolution which is (was?) the hook responsible to give content width

Yeah totally agree, this is the only logic that i recovered from this hook.

---
> We shouldn't have to maintain the toripunks app which was developed by the toripunks team, but there was no real work on external dapps integration

Okay good to know.

---
> Most mobile screens are meant to be responsive and not have a "mobile screen" version, most components too, it was badly designed

Mhh yeah in this case i think i agree with you, but it's not the purpose of this PR i think, so i think i have to keep the conditions for mobile

---
> we could do that with the breadcrumbs but we need a way to navigate back on mobile native and desktop native

Yes I'm okay with that

---
> those are needed to automatically select and restricts networks, for example on a post details page, forceNetworkId is used to restrict to the network of the post, on the feed forceNetworkFeatures is used to restrict to networks that have a feed

Okay, can we consider that like "network optimizations" ? I don't think that we can refacto that.. because just for that, we need 3 props in the component but okay

n0izn0iz:
> If we create a new component and we replace it everywhere what is the problem for you ?

I don't understand the difference with what I ask then. What I don't want is NewScreenContainer and ScreenContainer existing together in the app

> Okay, can we consider that like "network optimizations" ?

No we NEED to filter networks with various constraints, we could merge the 3 in a something like `networkFilter` props though

I don't think we should spend too much time fixing something that isn't broken. IMO we should focus on adapting/simplifying the existing component. Starting with the scope we defined, making all screens responsive and taking the full width.
It's especially hard to understand the logic change that can happen if we create a new component and can't review a diff

clegirar:
> I don't understand the difference with what I ask then. What I don't want is NewScreenContainer and ScreenContainer existing together in the app

Yes no difference ^^ This is what i want to say.
Like i said It's WIP, this is the reason why there is 2 components, but it will not be like that that at the end, i just wanted to do an other component to test a specific screen, and check what is necessary in this component and what it's not, it will change obviously.

> No we NEED to filter networks with various constraints, we could merge the 3 in something like networkFilter props though

Okay okay, filter networks.

> I don't think we should spend too much time fixing something that isn't broken. IMO we should focus on adapting/simplifying the existing component. Starting with the scope we defined, making all screens responsive and taking the full width.

Just adapting this component with full width, is just adding complexity IMO. But if we don't have a lot of time to refacto/simplify it, it's okay, but this is a component that is used at the base of each screen, so it doesn't have to be a thing that we never change. But as you want, you have the final word 👍

I can propose to do that after changes of font family and full width screen everywhere.

Thanks for the discussion and giving me informations.

n0izn0iz:
> Just adapting this component with full width, is just adding complexity IMO

It already supports it, like in Marketplace screen, no?

clegirar:
> It already supports it, like in Marketplace screen, no?

It's `isLarge` props that makes the screen "full width" and all children let the possibility to do it (`width: "100%"` style props mainly), so it's not the good name props, and i think you have at least 2 possibilities to have "full width" screens but it's finally not the purpose of this PR.
So to do that for `FeedScreen` we have to change some styles from his children, but i'll change that and propose  something tonight.

hthieu1110:
@clegirar as I can see, according to feedbacks from Norman, this PR is reduced to just handled the fullWithWindow and no introduce any new Component ?

hthieu1110 (approved):
LGTM

clegirar:
Yep exactly, i'll refacto it in a second step 👍

MikaelVallenet (commented):
i saw this:

before:
image


after:
image

Looks weird to me wdyt

clegirar:
Yeah saw that too, but it's not the same screen, so i want do that in an other PR 👍

MikaelVallenet (approved):

WaDadidou:
The width is wrong in mobile mode. Our layouting is like a labyrinth. Be careful to the side effects 🤕 

On the PR: 
image

On main: 
image

clegirar:
Oh yeah, the mobile view, in my head i was like "no need to handle mobile version" but yeah you right, will fix that 👍

MikaelVallenet (dismissed):

n0izn0iz (commented):
Screenshot 2024-12-16 at 15 00 57
the navbar should take the full width and is in scope of this PR imo

packages/components/socialFeed/NewsFeed/NewsFeed.tsx
```diff
@@ -233,7 +228,11 @@ export const NewsFeed: React.FC<NewsFeedProps> = ({
         }
         keyExtractor={(post) => post.id}
         onScroll={scrollHandler}
-        contentContainerStyle={contentCStyle}
+        contentContainerStyle={{
```
hthieu1110, line 231: This is just personal style but I'd write like this to avoid style undefined. (just my opinion, dont need to change your code)

isMobile ? {
    alignItems: "center",
    width: 100%
} : {
    alignSelf: center,
    width
}

packages/hooks/useMaxResolution.ts
```diff
@@ -22,12 +20,13 @@ export const useMaxResolution = ({
   isLarge = false,
 } = {}) => {
   const { width: windowWidth, height: windowHeight } = useWindowDimensions();
-  const { isSidebarExpanded } = useSidebar();
   const isMobile = useIsMobile();
+
+  // If we have a different width when sidebar is expanded and when it's not, the sidebar will be laggy on certain screens (like FeedScreen)
+  // So this calcul find the bigger width to have the same width no matter the sidebar state
   const contentWidth = useMemo(
```
hthieu1110, line 27: IMHO, useMemo does not gain much in perf. But we can let it, there is no problem.

hthieu1110 (dismissed):
LGTM

packages/hooks/useMaxResolution.ts
clegirar: I didn't change the `useMemo` i just explained that if the width change between sidebars states, it will be laggy 👍

clegirar:
> the navbar should take the full width and is in scope of this PR imo 

It's more a proposition than a definition of this task, but i can try it if you want obviously :)

n0izn0iz (approved):
LGTM

hthieu1110 (approved):
LGTM

PR #1455, opened 2024-12-11
fix: use most recentday instead of yesterday

hthieu1110:
This aims at fixing the total rewards calculation in indexer. For now, we get the last data by calculating yesterday from today date => We have to query the most recent date instead of yesterday.


omarsy (approved):
LGTM

MikaelVallenet (approved):

PR #1456, opened 2024-12-16
feat: allow to make article proposal

n0izn0iz:
Screenshot 2024-12-16 at 14 28 07

When you have a wallet connected that is member of DAOs, you can select a DAO to post with

### DIFF
--- a/packages/components/socialFeed/RichText/RichText.tsx
+++ b/packages/components/socialFeed/RichText/RichText.tsx
@@ -28,6 +28,7 @@ export const RichText: React.FC<RichTextProps> = ({
   loading,
   isPostConsultation,
   initialValue,
+  publishText = "Publish",
 }) => {
   const { width: windowWidth, height: windowHeight } = useWindowDimensions();
   const richText = useRef(null);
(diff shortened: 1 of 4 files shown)


MikaelVallenet (dismissed):

WaDadidou:
It's LGTM for me, except for one issue (fixed here: https://github.com/TERITORI/teritori-dapp/pull/1456/commits/91d47794b731f2f705b6c5b8861f3dc0cd1a45d4)

The issue:
If I choose a DAO wallet, then I select another wallet from Keplr, `DAOSelector` is not resetted, so it keep the selected DAO wallet, and "Propose" in the button
This state bellow must be impossible:
image

WaDadidou (approved):
It's LGTM for me, except for one issue (fixed here: https://github.com/TERITORI/teritori-dapp/pull/1456/commits/91d47794b731f2f705b6c5b8861f3dc0cd1a45d4)

The issue:
If I choose a DAO wallet from our front, then I select another wallet from Keplr, `DAOSelector` is not resetted, so it keep the selected DAO wallet, and "Propose" in the button
This state bellow must be impossible:
image

MikaelVallenet (approved):

MikaelVallenet:
it would also be nice (in another PR since it's not the scope) but to use the start of the article to make a preview subtitle and let it optional to write a custom preview

PR #1468, opened 2024-12-29
chore(gno): refacto dao modules

n0izn0iz:
- rename `dao_roles_voting_group` into `dao_voting_roles`
- remove the `dao_roles_group` struct and package
- only allow to pass a single resource in `VotingPowerAtHeight`
- allow to pass a single ressource in `TotalVotingPowerAtHeight`
- allow only a single message in `dao_proposal_single`


n0izn0iz:
closing in favor of replacing with daokit

PR #1469, opened 2024-12-30
feat(gno/orgs): save & display kind of organizations

MikaelVallenet:
an organization could be based on different kinds:
- roles
- memberships
- tokens
- ...

This information should be saved into contracts & retrievable to be able to adapt business logic depending on the kind of the org.

The PR scope is to save it and display it in UPP

-----------------------------------

## View of UPP org after PR

image

&&

image


gno/r/dao_realm/dao_realm.gno
```diff
@@ -94,7 +94,7 @@ func RegisterSelf() {
 		panic("already registered")
 	}
 
-	dao_registry.Register(func() dao_interfaces.IDAOCore { return daoCore }, "DAO Realm", "Default testing DAO", "https://images.unsplash.com/photo-1477959858617-67f85cf4f1df?w=1080&fit=max")
+	dao_registry.Register(func() dao_interfaces.IDAOCore { return daoCore }, "DAO Realm", "Default testing DAO", "memberships based organization", "https://images.unsplash.com/photo-1477959858617-67f85cf4f1df?w=1080&fit=max")
```
n0izn0iz, line 97: The registry should only be mapping from realm pkgpath to IDAOCore, the additional fields were there when the profile realm didn't exist but they should be removed actually, this is a review mistake by me when we moved to profile realm for daos

I think we should handle the "dao kind" by inspecting the dao voting module, which already has a `.Info()` api

n0izn0iz (changes requested):

gno/r/dao_realm/dao_realm.gno
MikaelVallenet: So when listing (for the screens w/ all daos) should i fetch profile realm (name, bio & avatar) for each field of each DAO ?

I did what you suggest to handle the dao kind by inspecting the dao voting module using the Info().String() endpoint

n0izn0iz: there is two option there:
- query the profile realm in the registry list function and enrich the response within gno
- use the useUserInfo hook like we do for users, it already supports gno dao if I'm not mistaken

n0izn0iz (approved):
LGTM

WaDadidou (approved):
LGTM :)

PR #1470, opened 2024-12-31
feat(rakki): Add RAKKi to sidebar

WaDadidou:
### Adding RAKKi to the sidebar
image

### Adding notch color customization depending on the sidebar item id
https://github.com/TERITORI/teritori-dapp/blob/b82fc13df32c7758b96743460c1a14efc00f2618/packages/components/navigation/components/SideNotch.tsx#L10
Need that to make RAKKi yellow

### DIFF
--- a/packages/components/navigation/components/SideNotch.tsx
+++ b/packages/components/navigation/components/SideNotch.tsx
@@ -1,26 +1,33 @@
 import React from "react";
-import { StyleSheet, View, ViewStyle } from "react-native";
+import { View, ViewStyle } from "react-native";
 
 import SideNotchSVG from "../../../../assets/sidebar/side-notch.svg";
 import { SVG } from "../../SVG";
 
-export const SideNotch: React.FC<{ style?: ViewStyle }> = ({ style }) => {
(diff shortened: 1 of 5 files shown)


hthieu1110 (approved):
LGTM

packages/components/navigation/components/SideNotch.tsx
```diff
@@ -1,26 +1,33 @@
 import React from "react";
-import { StyleSheet, View, ViewStyle } from "react-native";
+import { View, ViewStyle } from "react-native";
 
 import SideNotchSVG from "../../../../assets/sidebar/side-notch.svg";
```
hthieu1110, line 4: should we use alias import to avoid long ../ :D

MikaelVallenet (approved):

packages/components/navigation/components/SideNotch.tsx
WaDadidou: We can use `@/` instead of `../../../..` yes.
I didn't change it here to limit the diffs

PR #1471, opened 2025-01-02
WIP: fix: refacto react-hook-form logic for roles based organizations

clegirar:
Start a refacto about react-hook-form logic for roles based organizations.
The goal is to understand better the react-hook-form flow because it's a lot of abstractions. You can give your opinion if you find it more appropriate and understandable.

TODO:
- [ ] Re-check naming, especially for zod objects and types
- [ ] Re-check logic
- [ ] Test the process a lot

Can't refacto `ConfigureVotingFormData` and `OrganizationData` because it's generic between all organizations, but could be cool to adapt it.

### DIFF
--- a/packages/utils/types/organizations.ts
+++ b/packages/utils/types/organizations.ts
@@ -1,3 +1,5 @@
+import { z } from "zod";
+
 import { NSAvailability } from "./tns";
 
 export enum DaoType {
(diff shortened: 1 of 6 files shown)


packages/utils/types/organizations.ts
```diff
@@ -27,6 +29,35 @@ export type ConfigureVotingFormType = {
   minutes: string;
 };
 
+export const ZodRoleObject = z.object({
```
n0izn0iz, line 32: please don't capitalize zod schema as they are objects
```suggestion
export const zodRoleObject = z.object({
```

n0izn0iz (changes requested):

WaDadidou:
Hello, I've added a custom error message to `min(1)` usages to avoid differences between:
"submit with no value in field" 
image
and "write value and delete written value in field"
image

I also changed the value for `DEFAULT_FORM_ERRORS.required` to have the same error as default RHF's "required" error.

Commit: https://github.com/TERITORI/teritori-dapp/pull/1471/commits/e63f41034ec77086a3abd47a0cd64cc92c72057d

PR #1476, opened 2025-01-04
fix: text color for TextInputCustom

clegirar:


packages/components/inputs/TextInputCustom.tsx
```diff
@@ -353,6 +353,7 @@ const styles = StyleSheet.create({
   },
   textInput: {
     ...fontRegular14,
+    color: secondaryColor,
```
WaDadidou, line 356: Why not using `theme.textColor` ?

clegirar: Yeah found it better too, just wanted to revert but okay agree

WaDadidou (dismissed):

PR #1487, opened 2025-01-07
feat(organizations): new contract explo w/ conditions

MikaelVallenet:
use new pkg #1479 

### Step 1 - Naive initialization
goal: build a dao mostly hardcoded that use the new pkg and new way of processing vote
requirements:
- [ ] Should have a list of members w/ Roles & resources linked to condition at initialization
- [ ] Should have respectiv method to list roles, members & resources w/ conditions
- [ ] Should have a render gnoweb that show all the things mentioned above

### Step 2 - Naive voting & proposal
goal: allow people from dao to create proposal & use the conditions to process the voting & execution
- [ ] Should have a method to vote that precise the proposalId & the vote content
- [ ] Should have a way to render current state of a proposal
- [ ] Should list proposal on gnoweb render & be able to click on each to have more details about the status of each condition ...etc

### Step 3 - Use the previous step to think of how we will build the "final" version
:warning: I want to reflect on previous steps & maybe sync w/ crew to define how should be final version


gno/p/dao2/dao2.gno
```diff
@@ -0,0 +1,101 @@
+package dao2
+
```
MikaelVallenet, line 2: remove this

gno/p/dao2/gno.mod
```diff
@@ -0,0 +1 @@
+module gno.land/p/teritori/dao2
```
MikaelVallenet, line 1: remove this too

MikaelVallenet:
ADD unit test & link to vote or execute when the proposal is ready to be execute
Add a link to add a proposal from the UI

gno/p/daokit/daokit.gno
```diff
@@ -0,0 +1,180 @@
...
+		panic("invalid vote")
+	}
+
+	e := &daocond.EventVote{
+		VoterID: voter.String(),
+		Vote:    daocond.Vote(vote),
+	}
+
+	proposal.state.HandleEvent(e, proposal.votes)
+	proposal.votes[voter.String()] = daocond.Vote(vote)
```
Villaquiranm, line 97: maybe a bad idea, but what do you think of handling the vote record inside the p realm ? like that is not resposability of the realm caller to register that vote.

If the vote is not registered it could lead to bugs right ? (for example we could be able to vote multiple times )

gno/p/daokit/daokit.gno
```diff
@@ -0,0 +1,180 @@
...
+	if !d.MemberModule.IsMember(executor.String()) {
+		panic("executor is not a member")
+	}
+
+	proposal := d.ProposalModule.getProposal(proposalID)
+	if proposal == nil {
+		panic("proposal not found")
+	}
+
+	if proposal.status != ProposalStatusOpen {
```
Villaquiranm, line 112: Something to discuss during our point:
State and status seems like too close one to another maybe we need to rename one or another

gno/p/daokit/proposals.gno
```diff
@@ -0,0 +1,101 @@
...
+	value, ok := p.proposals.Get(seqid.ID(id).String())
+	if !ok {
+		return nil
+	}
+	proposal := value.(*Proposal)
+	return proposal
+}
+
+func (p *Proposal) updateStatus() {
+	e := p.state.Eval(p.votes)
```
Villaquiranm, line 90: maybe rename it to something like conditionsAreMeet ?

gno/p/daokit/proposals.gno
```diff
@@ -0,0 +1,101 @@
...
+	if p.status == ProposalStatusOpen && e {
+		p.status = ProposalStatusPassed
+	}
```
Villaquiranm, line 93:
```suggestion
	conditionsAreMet := p.state.Eval(p.votes)
	if p.status == ProposalStatusOpen && conditionsAreMet {
		p.status = ProposalStatusPassed
	}
```

Villaquiranm: maybe ?

gno/p/daokit/daokit.gno
MikaelVallenet: I disagree, i don't see advantage to let the realm handle the voting records
I think we should have minimal realm and abstract all dao related logic in the daokit package

gno/p/daokit/daokit.gno
MikaelVallenet: condition state

PR #1489, opened 2025-01-09
feat: animation when you buy rakki tickets

clegirar:
There is a preview [here](https://www.loom.com/share/1ff4d954a6644bfda60e41ca442cdbc4) !

In the code we have 3 animations of confettis, because it's a `ScrollView` and the height is higher than a screen, so i put 3 animations in the entire `ScrollView` to see the animation wherever you are in the View.


WaDadidou:
Maybe it's a detail, but a space is scrollable at the bottom, on a small zoom level 
image
I think we could simplify the confettis handling by just force scrolling to the top, then show the confettis

WaDadidou:
The zoom level chanages the position of the animation. And it adds an extra vertical space at the bottom
image


What do you think about this adjustment ? https://github.com/WaDadidou/teritori-dapp/commit/91e7d9e39cab5c75ec7917e9d531ec6b330506aa#diff-8e2c5b8031f032da60995a2ec1bed47f31b3db49f0ce3de39821c1d7f26382cfR103

- Only 2 LottieAnimation
- Centered regardless of the zoom level

image

clegirar:
Screenshot 2025-01-14 at 09 59 12

Got something like this, how do you find it @WaDadidou ?

And i think i found a better calcul without get `screenContainerContentHeight` 👍

WaDadidou (approved):
Ah yes yes, better :) LGTM

hthieu1110 (approved):
LGTM

PR #1491, opened 2025-01-09
feat: enable auto-compaction of js code

n0izn0iz:


clegirar (approved):

n0izn0iz:
this option only works on markdown

PR #1499, opened 2025-02-05
fix: Update upload-artifact to v4 because deprecated v3

WaDadidou:


package.json
```diff
@@ -96,20 +96,20 @@
...
-    "expo": "~50.0.20",
-    "expo-av": "~13.10.6",
-    "expo-blur": "~13.0.3",
-    "expo-clipboard": "~5.0.1",
-    "expo-document-picker": "~11.10.1",
-    "expo-file-system": "~16.0.9",
-    "expo-font": "~11.10.3",
-    "expo-image-picker": "^14.7.1",
-    "expo-linear-gradient": "~12.7.2",
+    "expo": "latest",
```
n0izn0iz, line 99: why do you need this? I think only syncing explo-blur should be okay
I any case we shouldn't change the expo version without doing the proper update walkthrough

n0izn0iz (changes requested):

package.json
WaDadidou: It's a mistake

n0izn0iz: I strongly suggest to revert all the js change and start fresh, you shouldn't need to update any typescript code

WaDadidou: Yes T_T

PR #1503, opened 2025-02-14
feat: refacto daokit

n0izn0iz:
- Changed how you declare message handlers (much less boilerplate)
- Removed members, roles and render stuff from `daokit`. Now `daokit` is only responsible for proposals and resources (handler + condition). it doesn't do any check on caller and is unsafe to expose
- Added `basedao` package that adds members, roles, profile and render on top of `daokit`, this instantiates the default handlers if needed and add caller-is-member guards around vote, propose and execute
- Adapted `dao_realm` and `govdao` to use `basedao`
- Did some minor type changes (example: uint32 for counts so it can be represented with a javascript Number)
- Removed auth checks in `role_manager` and adapt callsites


gno/p/basedao/basedao.gno
```diff
@@ -0,0 +1,118 @@
...
+
+		if conf.SetProfileString != nil {
+			dao.Core.SetResource(&daokit.Resource{
+				Handler:   NewEditProfileHandler(conf.SetProfileString, []string{"DisplayName", "Bio", "Avatar"}),
+				Condition: conf.InitialCondition,
+			})
+		}
+
+		// register management handlers
+		for _, m := range []daokit.MessageHandler{
```
Villaquiranm, line 75: shouldn't this be dao_interfaces.MessageHandler ?
```suggestion
		for _, m := range []dao_interfaces.MessageHandler{
```

MikaelVallenet (approved):
LGTM :fire:

gno/p/basedao/basedao.gno
n0izn0iz: dao_interfaces is the old framework, it's still used by social feed and front (thus e2e tests) so I kept it

PR #1506, opened 2025-02-16
feat: Claim tiers functionality

Villaquiranm:


gno/r/govdao/seed.gno
```diff
@@ -0,0 +1,104 @@
...
+	"stefann-01":           "T2",
+	"Kouteki":              "T2",
+	"MikaelVallenet":       "T2",
+	"MalekLahbib":          "T2",
+	"schollz":              "T2",
+	"agherasie":            "T2",
+	"kristovatlas":         "T2",
+	"n2p5":                 "T2",
+	"faddat":               "T2",
+	"x1unix":               "T3",
```
n0izn0iz, line 44: we should remove T3 here as they must be elected with invitations
sorry for the bad initial request

(see Mikael's comment in `govdao.gno`)

n0izn0iz (changes requested):

gno/r/govdao/seed.gno
Villaquiranm: @n0izn0iz done here ad4fdefeb5f8c4c3e5b3a8e82c5e6b288772822b
and for the generate script:
https://github.com/samouraiworld/gnolove/pull/9/commits/f5bee3d2ece2f78d653f6f4f3dbf09445e59ae08

PR #1513, opened 2025-02-18
feat(gno/daokit): "resource management" handlers

Villaquiranm:


gno/p/daokit/resources.gno
```diff
@@ -32,3 +38,34 @@ func (r *resourcesStore) getResource(name string) *Resource {
...
+		if !ok {
+			panic(errors.New("invalid payload type"))
+		}
+
+		if getResource(payload.Handler.Type()) != nil {
+			panic("ressource " + payload.Handler.Type() + " already exists")
+		}
+		r.setResource(payload)
+	})
+}
```
n0izn0iz, line 56:
```suggestion
func NewSetResourceHandler(r *resourcesStore) MessageHandler {
	return NewMessageHandler(MsgSetResourceKind, func(ipayload interface{}) {
		payload, ok := ipayload.(Resource)
		if !ok {
			panic(errors.New("invalid payload type"))
		}
		r.setResource(payload)
	})
}
```

n0izn0iz (changes requested):

n0izn0iz (commented):
when we overwrite or remove a resource, we should probably close/remove the proposals that use that resource

gno/p/daokit/resources.gno
Villaquiranm: done e31f9e352b1d2a1c8c962ab4d9f34fcce429dde3

PR #1516, opened 2025-02-22
fix(articles): Remove max width in article creation

WaDadidou:
To complete the work of @clegirar about max screens max width, I adjusted the max width in the Article creation's screen.

You can compare:

Before: https://app.teritori.com/feed/new
After: _TODO_

After:
image


hthieu1110 (approved):
LGTM

PR #1523, opened 2025-07-19
fix: Protobuf generation

WaDadidou:
`make generate` and `make buf.lint` give this error:
```
buf generate api
Failure: failed to build input "api" because it is contained by module at path "." specified in your configuration, you must provide the workspace or module as the input, and filter to this path using --path
make: *** [Makefile:68 : generate.protobuf] Erreur 1
```

So I fixed it and ran `make generate`

___
- `--path` flag added:
   - https://github.com/TERITORI/teritori-dapp/blob/fix-buf-generate/Makefile#L68
   - https://github.com/TERITORI/teritori-dapp/blob/fix-buf-generate/Makefile#L92
- `out` modfied to satisfy `--path` flag: https://github.com/TERITORI/teritori-dapp/blob/fix-buf-generate/buf.gen.yaml#L8
- `DEFAULT` replaced because deprecated: https://github.com/TERITORI/teritori-dapp/blob/fix-buf-generate/buf.yaml#L7


packages/api/api/dao/v1/dao.ts
```diff
@@ -0,0 +1,701 @@
+/* eslint-disable */
```
n0izn0iz, line 1: there shouldn't be files in api/api

n0izn0iz (changes requested):

packages/api/api/dao/v1/dao.ts
WaDadidou: Fixed.
- Added required`path` flag: https://github.com/TERITORI/teritori-dapp/pull/1523/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R68
- Removed `api` level from `out`: https://github.com/TERITORI/teritori-dapp/pull/1523/files#diff-10c91cb1168fcd0255be771c0387060b3b39acff43878dd86aaf468b10910f2aR8

Hm... Is it ok?

n0izn0iz:
closing in favor of https://github.com/TERITORI/teritori-dapp/pull/1524

